PR #1, opened 2021-07-09
Embedding chainspec for vitalam chain

mattdean-digicatapult:
So this changes allows us to embed the genesis definition for our production chain into the node itself. It works by adding a "staging_testnet" chain which then has defined in it our initial validators, balances etc. I have then exported the chain spec using:

```sh
vitalam-node build-spec --disable-default-bootnode --chain=vitalam-staging > vitalam.json
```

This is then editted to change some of the chain parameters before encoding into a raw chainspec:

```sh
vitalam-node build-spec --disable-default-bootnode --raw --chain=./vitalam.json > vitalamRaw.json
```

It is this raw chain spec included in this PR. Now a few notes:

* the approach is based on how https://github.com/paritytech/polkadot/ seems to do things
* from a security PoV this is fine as we are only embedding public identities of the initial validators
* bootnodes can be added at CLI for our private chain deployment

I think this will work but it's somewhat hard to test until I update some other bits.


jonmattgray (approved):

node/src/chain_spec.rs
```diff
@@ -7,6 +7,7 @@ use vitalam_node_runtime::{
     AccountId, AuraConfig, BalancesConfig, GenesisConfig, GrandpaConfig, Signature, SudoConfig, SystemConfig,
     WASM_BINARY,
 };
+const DEFAULT_PROTOCOL_ID: &str = "vam";
```
callumhyland-dc, line 10: 🌟

callumhyland-dc (approved):

PR #3, opened 2021-07-09
VIT-36 Add destroyed_at to Token

callumhyland-dc:
(sorry for diff churn 🌪️ )

- [x] Adds `created_at` and `destroyed_at` to Token type.
- [x] Updates tests
- [x] Implement `on_runtime_upgrade`
- [x] Update weights


pallets/simple-nft/src/lib.rs
```diff
@@ -56,7 +72,48 @@ pub mod pallet {
...
+                    },
+                    metadata: old_token.metadata,
+                    parents: old_token.parents,
+                    children: old_token.children,
+                };
+
+                put_storage_value(b"SimpleNFTModule", b"TokensById", &key_hash, new_token);
+            }
+
+            0
```
mattdean-digicatapult, line 114: I think we should make an attempt at getting this right rather than just returning `0`. Have you seen any indication of best practice for coming up with the correct weight?

pallets/simple-nft/src/lib.rs
```diff
@@ -56,7 +72,48 @@ pub mod pallet {
     pub struct Pallet<T>(PhantomData<T>);
 
     #[pallet::hooks]
-    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
mattdean-digicatapult, line 76: This looks good but is untested. Have you seen any examples of migration tests you could look at for reference?

callumhyland-dc:
👍

mattdean-digicatapult (approved):
Approving now we have the thumbs up from @callumhyland-dc

PR #4, opened 2021-07-16
VIT-115 Limit node connections and access to chain data

jonmattgray:
We want to limit which nodes connect to other nodes on the chain to stop unknown nodes from accessing the chain and reading chain data. Substrate provide the node-authorization pallet for creating and managing a permissioned network of nodes.

The file changes cover the following steps:
- [x] Add [node-authorization](https://docs.rs/pallet-node-authorization/3.0.0/pallet_node_authorization/index.html) pallet.
- [x] Set pre-authorised well-known nodes in genesis config for each chain type.
```
dev
  - Alice

staging
  - stage (red/green/blue/bootnode)
  - stage2 (red/green/blue)
  - stage3 (red/green/blue)

local
  - Alice
  - Bob
  - Charlie 
  - Eve
```
- [x] Bump node version

The process for adding genesis config to a live chain (in this case adding an initial set of permissioned well-known nodes) is documented [here](https://digicatapult.atlassian.net/wiki/spaces/EN/pages/1736015886/Adding+genesis+config+to+a+live+chain).
The process for adding a new node, either as a well-known node or a sub-node, to an already permissioned chain with an existing set of well-known nodes is documented [here](https://digicatapult.atlassian.net/wiki/spaces/EN/pages/1722253361/Node+Authentication#Example).

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -157,8 +157,11 @@ pub fn vitalam_staging_testnet_config() -> Result<ChainSpec, String> {
     ];
 
     let endowed_accounts: Vec<AccountId> = vec![
+        // 5Fpk38Xk2eixKwtNRnUn1YTcUuKbgSmF8G2wgRoqGUmgtG1o
         hex!["a64ad684438ae298d875f66e24e4f98f2d4689e008d431534f5c6adffd1ea26b"].into(),
+        // 5DtQJtJurJbniRA1LaFu9JgoL5Myi4C29EBfqtoJpcoEydLp
         hex!["509cf0e9e9c49855cf888f9af6bd481174399d6fdbe4d43041e6ed8a960b6a2c"].into(),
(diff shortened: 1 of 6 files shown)


node/src/chain_spec.rs
```diff
@@ -219,6 +323,22 @@ pub fn local_testnet_config() -> Result<ChainSpec, String> {
                     get_account_id_from_seed::<sr25519::Public>("Eve//stash"),
                     get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
                 ],
+                Some(NodeAuthorizationConfig {
```
mattdean-digicatapult, line 326: I think we need to add Eve to this list in order for part of the demo to work?

mattdean-digicatapult (commented):
Looks good. Last question is whether this causes issues for the demo?

mattdean-digicatapult (approved):

PR #8, opened 2021-11-10
Multiple chain metadata values

jonmattgray:
Having multiple metadata values will allow us to separate out the JSON blob from the base64 encoded content file in the demo. This will allow someone to understand the meaning of a token without needing to download the attached file.

- [x] Multiple metadata (key, value) pairs against a token
- [x] Update tests
- [x] Remove old migration 
- [x] Configurable and enforced maximum number of metadata items supported by the pallet
- [x] Update codeowners
- [x] Minor version bump -> `2.1.0`
- [x] Update README
- [x] Regenerate weights for run_process


pallets/simple-nft/src/lib.rs
```diff
@@ -109,7 +111,7 @@ pub mod pallet {
         pub(super) fn run_process(
```
mattdean-digicatapult, line 111: need to regenerate weights for `run_process`. I think this is described inthe readme 🤔

pallets/simple-nft/src/tests.rs
```diff
@@ -320,51 +344,77 @@ fn it_fails_for_destroying_multiple_tokens_as_other_signer() {
...
         assert_eq!(token_1, SimpleNFTModule::tokens_by_id(1));
-        // asset old token hasn't changed
+        // assert old token hasn't changed
         assert_eq!(token_2, SimpleNFTModule::tokens_by_id(2));
     });
 }
+
+
+#[test]
+fn it_fails_for_creating_single_token_with_too_many_metadata_items() {
```
mattdean-digicatapult, line 403: Need another test covering multiple metadata values on a token and perhaps one where multiple outputs have different numbers of valid metadata items

runtime/src/lib.rs
```diff
@@ -272,12 +272,18 @@ impl pallet_node_authorization::Config for Runtime {
...
+parameter_types! {
+    pub const MaxMetadataCount: u32 = 16;
+}
+
 /// Configure the template pallet in pallets/simple-nft.
 impl pallet_simple_nft::Config for Runtime {
     type Event = Event;
     type TokenId = u128;
-    type TokenMetadata = Hash;
+    type TokenMetadataKey = Vec<u8>;
```
mattdean-digicatapult, line 283: I wonder if this would be better as a length limited array, `[u8; 32]` say

mattdean-digicatapult (changes requested):
Few minor issues. Looking good

pallets/simple-nft/src/tests.rs
jonmattgray: Added a multiple metadata values on a single token test and I think I've added the second one?

jonmattgray: https://github.com/digicatapult/vitalam-node/pull/8/commits/305ff62fd14cc0ac3e0a7e6cf5a5be35589dd5a9

runtime/src/lib.rs
jonmattgray: changed, still `u64` for tests

jonmattgray: and updated API to handle padding

pallets/simple-nft/src/lib.rs
jonmattgray: done

darrylmorton-digicatapult (approved):

pallets/simple-nft/src/weights.rs
```diff
@@ -27,18 +27,16 @@ pub trait WeightInfo {
...
+	fn run_process(i: usize, o: usize) -> Weight {
+		(104_267_000 as Weight)
+			// Standard Error: 6_647_000
+			.saturating_add((13_768_000 as Weight).saturating_mul(i as Weight))
+			.saturating_add(T::DbWeight::get().reads(1 as Weight))
+			.saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(i as Weight)))
+			.saturating_add(T::DbWeight::get().writes(1 as Weight))
+			.saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(i as Weight)))
+			.saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(o as Weight)))
+	}
```
mattdean-digicatapult, line 39: This is suspicious. Specifically that the base weight has gone up by so much and there is now no element based on processing the each output like ` .saturating_add((8_875_000 as Weight).saturating_mul(o as Weight))`. I think we can let this pass, but we should look into improving these as time goes on

mattdean-digicatapult (approved):
LGTM with one comment for the future. Can you create a ticket to re-evaluate our weights and benchmarks once the process abstraction is done?

PR #10, opened 2021-11-16
on-chain members list

darrylmorton-digicatapult:
on-chain pallet-membership added


mattdean-digicatapult (commented):
This looks good, but I'd like to hold off approving until the integration with the API is tested

mattdean-digicatapult (approved):

PR #12, opened 2021-11-30
Multiple token roles

jonmattgray:
Represent multiple roles on a token beyond `owner` so we can more easily understand which accounts it's allowable to transfer a token to. In the future this will allow multiple parties to make the next statement on an asset (token) without having to transact back and forth.

`owner: AccountId` -> `roles: BTreeMap<RoleKey, AccountId>`

At runtime, `RoleKey` has type of `Role` which represents an enumeration of the sensible roles (agreed with Luca) we anticipate for the project.
```
pub enum Role {
    Admin, //Default
    ManufacturingEngineer,
    ProcurementBuyer,
    ProcurementPlanner,
    Supplier,
}
```

For now, `run_process` checks that the sender `AccountId` matches the `AccountId` for the default role (`Admin`) of all `input` tokens, more nuance will be added to this in the future.

- [x] version -> `2.4.0`
- [x] `owner: AccountId` ->  `roles: BTreeMap<RoleKey, AccountId>`
- [x] add RoleKey + Role enum
- [x] validate that at least the default role is set on every token 
- [x] update tests to cover new enum
- [x] benchmarking
- [x] README

### DIFF
--- a/pallets/simple-nft/src/mock.rs
+++ b/pallets/simple-nft/src/mock.rs
@@ -68,6 +68,7 @@ parameter_types! {
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, Ord, PartialOrd)]
 pub enum Role {
     Admin,
+    NotAdmin,
 }
 
 impl Default for Role {
--- a/pallets/simple-nft/src/mock.rs
+++ b/pallets/simple-nft/src/mock.rs
@@ -67,12 +67,12 @@ parameter_types! {
 
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, Ord, PartialOrd)]
 pub enum Role {
-    Owner,
+    Admin,
 }
 
 impl Default for Role {
     fn default() -> Self {
-        Role::Owner
+        Role::Admin
     }
 }
 
--- a/pallets/simple-nft/src/mock.rs
+++ b/pallets/simple-nft/src/mock.rs
@@ -65,6 +65,17 @@ parameter_types! {
     pub const MaxMetadataCount: u32 = 3;
 }
 
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, Ord, PartialOrd)]
+pub enum Role {
+    Owner,
+}
+
+impl Default for Role {
+    fn default() -> Self {
+        Role::Owner
+    }
+}
+
 #[derive(Encode, Decode, Clone, PartialEq, Debug, Eq)]
 pub enum MetadataValue {
     File(Hash),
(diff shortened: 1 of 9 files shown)


pallets/simple-nft/src/mock.rs
```diff
@@ -65,6 +65,17 @@ parameter_types! {
     pub const MaxMetadataCount: u32 = 3;
 }
 
+#[derive(Encode, Decode, Clone, PartialEq, Debug, Eq, Ord, PartialOrd)]
+pub enum Role {
+    Admin,
+}
```
mattdean-digicatapult, line 71: I think we should have a test using multiple roles.

pallets/simple-nft/src/weights.rs
```diff
@@ -28,9 +28,11 @@ pub trait WeightInfo {
...
-            // Standard Error: 6_647_000
-            .saturating_add((13_768_000 as Weight).saturating_mul(i as Weight))
+        (0 as Weight)
+            // Standard Error: 3_008_000
+            .saturating_add((14_764_000 as Weight).saturating_mul(i as Weight))
+            // Standard Error: 3_008_000
+            .saturating_add((10_806_000 as Weight).saturating_mul(o as Weight))
             .saturating_add(T::DbWeight::get().reads(1 as Weight))
             .saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(i as Weight)))
             .saturating_add(T::DbWeight::get().writes(1 as Weight))
```
mattdean-digicatapult, line 38: I think we need to have a look at the benchmark here and see what we can / should change. I feel maybe we should be considering iterating both the number of metadata values and the number of roles

mattdean-digicatapult (changes requested):
Looking very close. MIssing some tests and I think we need to look at updating the benchmarks

pallets/simple-nft/src/weights.rs
jonmattgray: Based on our talk earlier we're not considering the number of metadata values or roles (yet!).

I've replaced these values with new ones from a 1000-repeat benchmark. The results can still _sometimes_ vary signficantly between sets of 1000-repeat runs, but most seem to fall close to the values I've committed.

pallets/simple-nft/src/mock.rs
jonmattgray: I've added another Role to this mock enum + added a test where more than one role is set.

mattdean-digicatapult: @jonmattgray I only see one test referring to the new role `NotAdmin` (`it_fails_for_destroying_single_token_as_incorrect_role`) which is clearly a failure case. Am I missing something?

jonmattgray: you're right I'll add a happy path one for multiple roles

jonmattgray: new test added

mattdean-digicatapult (approved):

PR #13, opened 2021-12-06
Token traceability

jonmattgray:
Adds a new `original_id` property to tokens so we can track a tokenised asset through the transaction graph without having to download metadata from IPFS. Lets us more easily construct the asset's total state.

`outputs` for `run_process` now takes an `output` struct with a new optional third argument `Some<u32>` which refers to an index of the `inputs` that the output should be (uniquely) assigned to. If a valid index is given, the new token's `original_id` is set to the `original_id` of the consumed input. If `None` is given, `original_id` is set to the new token ID i.e. a new asset conceptually.

- [x] version -> `2.5.0`
- [x] add `output` struct 
- [x] add optional `parent_index` to `outputs` for `run_process` 
- [x] validate parent indices are within bounds and not duplicate within one `run_process`
- [x] update tests + add new ones to check validation
- [x] benchmarking
- [x] README


pallets/simple-nft/src/weights.rs
```diff
@@ -28,16 +28,16 @@ pub trait WeightInfo {
...
+        (14_049_000 as Weight)
+            // Standard Error: 7_000
+            .saturating_add((12_902_000 as Weight).saturating_mul(i as Weight))
+            // Standard Error: 7_000
+            .saturating_add((5_897_000 as Weight).saturating_mul(o as Weight))
+            .saturating_add(T::DbWeight::get().reads(1 as Weight))
+            .saturating_add(T::DbWeight::get().reads((1 as Weight).saturating_mul(i as Weight)))
+            .saturating_add(T::DbWeight::get().writes(1 as Weight))
+            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(i as Weight)))
+            .saturating_add(T::DbWeight::get().writes((1 as Weight).saturating_mul(o as Weight)))
```
mattdean-digicatapult, line 40: Hang on did the weights really not change at all?

mattdean-digicatapult (changes requested):

pallets/simple-nft/src/weights.rs
jonmattgray: this is some churn as I forgot to `cargo fmt` after changing weights on the previous PR. I think in a distant call in 2021 we decided there was no need to update benchmarking for this PR

mattdean-digicatapult: The benchmarking doesn't need to change no, but the weights will likely be different so you just need to run the benchmarking suite again and re-integrate

pallets/simple-nft/src/lib.rs
```diff
@@ -118,6 +124,7 @@ pub mod pallet {
             outputs: Vec<(
                 BTreeMap<T::RoleKey, T::AccountId>,
                 BTreeMap<T::TokenMetadataKey, T::TokenMetadataValue>,
+                Option<u32>,
             )>,
```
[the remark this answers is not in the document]
jonmattgray: added `Output` struct

pallets/simple-nft/src/weights.rs
jonmattgray: ran again and got slightly different weights

dblane-digicatapult: Should we automate the benchmarking for outputting the weights?

mattdean-digicatapult: It would be nice but it's complicated because the benchmarking output is not a 1-2-1 with what ends up going here. I think we should have a backlog item for this

jonmattgray: https://digicatapult.atlassian.net/browse/IN-66

PR #19, opened 2022-01-12
Create minideb docker image

danielcooperxyz:
## Summary
A new docker file that builds and image using pre-compiled binaries.
- [x] Create dockerfile that uses minideb and precompiled binary
- [x] Modify the workflow to build docker after compiling the binary

## Motivation
Our docker builds are very slow, so we are considering a replacement with a simpler build that downloads pre-compiled artefacts.

## Describe alternatives you've considered
- ☑️ Docker build time and rust compilation optimisations
- ☑️ Caching for rust compilation

## Additional context
  ```
▶ docker run -it --rm vitalam-node:minideb          11:30:31
2022-01-12 11:30:36 VITALam Node    
2022-01-12 11:30:36 ✌️  version 2.5.1-7b7d8d4-x86_64-linux-gnu    
2022-01-12 11:30:36 ❤️  by Digital Catapult <https://www.digicatapult.org.uk>, 2020-2022    
2022-01-12 11:30:36 📋 Chain specification: VITALam    
2022-01-12 11:30:36 🏷 Node name: gifted-earth-7985    
2022-01-12 11:30:36 👤 Role: FULL    
2022-01-12 11:30:36 💾 Database: RocksDb at /root/.local/share/vitalam-node/chains/vitalam/db    
2022-01-12 11:30:36 ⛓  Native runtime: vitalam-node-201 (vitalam-node-1.tx1.au1)    
2022-01-12 11:30:36 🔨 Initializing Genesis block/state (state: 0xf6e9…35fc, header-hash: 0xc80e…dcc5)    
2022-01-12 11:30:36 👴 Loading GRANDPA authority set from genesis on what appears to be first startup.    
2022-01-12 11:30:36 ⏱  Loaded block-time = 6000 milliseconds from genesis on first-launch    
2022-01-12 11:30:36 🏷 Local node identity is: 12D3KooWJhzDQeWzRrRDZMZ9f5NtQgPFXptfhZuYj6Kd9g8786Ji    
2022-01-12 11:30:36 📦 Highest known block at #0    
2022-01-12 11:30:36 〽️ Prometheus server started at 127.0.0.1:9615    
2022-01-12 11:30:36 Listening for new connections on 127.0.0.1:9944.    
2022-01-12 11:30:41 💤 Idle (0 peers), best: #0 (0xc80e…dcc5), finalized #0 (0xc80e…dcc5), ⬇ 0 ⬆ 0    
2022-01-12 11:30:46 💤 Idle (0 peers), best: #0 (0xc80e…dcc5), finalized #0 (0xc80e…dcc5), ⬇ 0 ⬆ 0  
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,51 +1,27 @@
-# syntax=docker/dockerfile:1.2
+FROM bitnami/minideb:bullseye AS setup
 
-# build base that sets up common dependencies of the build
-FROM rust:alpine as build-base
+RUN apt-get update && apt-get install -y \
+  curl
+  #libgcc
+  # libstdc++
 
-RUN apk add --no-cache \
-  clang clang-dev clang-libs pkgconfig bearssl-dev git \
-  gcc make g++ linux-headers protobuf protobuf-dev musl-dev
+RUN mkdir /tmp/vitalam-node/
+WORKDIR /tmp/vitalam-node/
 
-RUN set -ex; \
-  wget https://github.com/gruntwork-io/fetch/releases/download/v0.4.2/fetch_linux_amd64 -P /; \
-  mv /fetch_linux_amd64 /fetch; \
-  chmod +x /fetch; \
-  /fetch --repo="https://github.com/mozilla/sccache" --tag="~>0.2.15" --release-asset="^sccache-v[0-9.]*-x86_64-unknown-linux-musl.tar.gz$" /; \
-  tar -xvf /sccache-v*-x86_64-unknown-linux-musl.tar.gz -C /; \
-  mv /sccache-v*-x86_64-unknown-linux-musl/sccache /sccache; \
-  rm -rf /sccache-v*-x86_64-unknown-linux-musl /sccache-v*-x86_64-unknown-linux-musl.tar.gz /fetch; \
-  chmod +x /sccache;
+RUN curl -L https://github.com/gruntwork-io/fetch/releases/download/v0.4.2/fetch_linux_amd64 --output ./fetch && chmod +x ./fetch
 
-WORKDIR /build
-ARG RUST_TOOLCHAIN=nightly-2021-11-09
-RUN rustup install $RUST_TOOLCHAIN && \
-  rustup target add wasm32-unknown-unknown --toolchain $RUST_TOOLCHAIN
+RUN ./fetch --repo="https://github.com/digicatapult/vitalam-node" --tag="v2.5.1" --release-asset="vitalam-node-v2.5.1-x86_64-unknown-linux-gnu.tar.gz" ./ \
+  && tar -xzf ./vitalam-node-v2.5.1-x86_64-unknown-linux-gnu.tar.gz 
 
-FROM build-base as build
+FROM bitnami/minideb:bullseye AS runtime
 
-COPY . .
-
-RUN --mount=type=cache,mode=0755,id=sccache,target=/cache \
-    --mount=type=tmpfs,target=/build/target \
-    PROTOC=$(which protoc) \
-    PROTOC_INCLUDE=/usr/include \
-    RUSTFLAGS=-Ctarget-feature=-crt-static \
-    RUSTC_WRAPPER=/sccache \
-    SCCACHE_DIR=/cache \
-    SCCACHE_IDLE_TIMEOUT=0 \
-    cargo build --release && \
-    cp /build/target/release/vitalam-node /vitalam-node
-
-# build the runtime image that will actually contain the final built executable
-
-FROM alpine:3.12 AS runtime
-
-RUN apk update
-RUN apk add libgcc libstdc++
+RUN apt-get update \
+  && apt-get install -y \
+  build-essential \
+  libgcc-10-dev
 
 RUN mkdir /vitalam-node /data
-COPY --from=build /vitalam-node /vitalam-node
+COPY --from=setup /tmp/vitalam-node /vitalam-node/
 
 WORKDIR /vitalam-node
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -1,56 +1,32 @@
...
-# build the runtime image that will actually contain the final built executable
-
-FROM alpine:3.12 AS runtime
-
-RUN apk update
-RUN apk add libgcc libstdc++
+RUN apt-get update \
+  && apt-get install -y \
+  build-essential \
+  libgcc-10-dev
```
mattdean-digicatapult, line 21: This seems like a lot for the runtime image. `build-essential` especially is a pretty meaty meta-dependency. What here to we actually need?

Dockerfile
```diff
@@ -1,56 +1,32 @@
-# syntax=docker/dockerfile:1.2
+FROM bitnami/minideb:bullseye AS setup
 
-# build base that sets up common dependencies of the build
-FROM rust:alpine as build-base
+RUN apt-get update && apt-get install -y \
```
mattdean-digicatapult, line 3: I think you should probably use `install-packages` (https://github.com/bitnami/minideb#why-use-minideb) for this to prevent excess metadata appearing in the docker images

Dockerfile
```diff
@@ -1,56 +1,32 @@
...
+  curl
+  #libgcc
+  # libstdc++
 
-RUN apk add --no-cache \
-  clang clang-dev clang-libs pkgconfig bearssl-dev git \
-  gcc make g++ linux-headers protobuf protobuf-dev musl-dev
+RUN mkdir /tmp/vitalam-node/
```
mattdean-digicatapult, line 8: I don't think you need a `mkdir` if you're then going to to a `WORKDIR`. It'll happen automatically I think

Dockerfile
```diff
@@ -1,56 +1,32 @@
...
-  mv /sccache-v*-x86_64-unknown-linux-musl/sccache /sccache; \
-  rm -rf /sccache-v*-x86_64-unknown-linux-musl /sccache-v*-x86_64-unknown-linux-musl.tar.gz /fetch; \
-  chmod +x /sccache;
+RUN curl -L https://github.com/gruntwork-io/fetch/releases/download/v0.4.2/fetch_linux_amd64 --output ./fetch && chmod +x ./fetch
 
-WORKDIR /build
-ARG RUST_TOOLCHAIN=nightly-2021-11-09
-RUN rustup install $RUST_TOOLCHAIN && \
-  rustup target add wasm32-unknown-unknown --toolchain $RUST_TOOLCHAIN
+RUN ./fetch --repo="https://github.com/digicatapult/vitalam-node" --tag="v2.5.1" --release-asset="vitalam-node-v2.5.1-x86_64-unknown-linux-gnu.tar.gz" ./ \
```
mattdean-digicatapult, line 13: Need a build-arg for the tag here. Release asset can also contain a wildcard I believe so you don't have to be specific there

mattdean-digicatapult (changes requested):

Dockerfile
danielcooperxyz: I've added the standard build arg env type situation. Please let me know if there is anything additional required.

Dockerfile
```diff
@@ -1,26 +1,22 @@
...
+RUN install_packages curl ca-certificates
 
-RUN mkdir /tmp/vitalam-node/
 WORKDIR /tmp/vitalam-node/
 
 RUN curl -L https://github.com/gruntwork-io/fetch/releases/download/v0.4.2/fetch_linux_amd64 --output ./fetch && chmod +x ./fetch
 
-RUN ./fetch --repo="https://github.com/digicatapult/vitalam-node" --tag="v2.5.1" --release-asset="vitalam-node-v2.5.1-x86_64-unknown-linux-gnu.tar.gz" ./ \
-  && tar -xzf ./vitalam-node-v2.5.1-x86_64-unknown-linux-gnu.tar.gz 
+ARG VITALAM_VERSION
```
mattdean-digicatapult, line 9: I'm not sure if this is a good idea but it might be sensible to provide a default of `latest` here. It'll make it easier for someone cloning the repo to build the docker image. That will require the use of a wildcard for the asset in the fetch call below. What do you think @dblane-digicatapult ?

danielcooperxyz:
```
 => ERROR [setup 5/5] RUN ./fetch --repo="https://github.com/digicatapult/vitalam-node" --tag="latest" --release-asse  0.9s
------
 > [setup 5/5] RUN ./fetch --repo="https://github.com/digicatapult/vitalam-node" --tag="latest" --release-asset="vitalam-node-latest-x86_64-unknown-linux-gnu.tar.gz" ./   && tar -xzf ./vitalam-node-latest-x86_64-unknown-linux-gnu.tar.gz:
#10 0.865 [fetch] time="2022-01-12T16:13:17Z" level=error msg="Could not find assets matching vitalam-node-latest-x86_64-unknown-linux-gnu.tar.gz in release latest\n"
------
```

mattdean-digicatapult: So you can use a regex for `--release-asset` (https://github.com/gruntwork-io/fetch#general-usage)

Dockerfile
```diff
@@ -1,56 +1,29 @@
...
 
 WORKDIR /vitalam-node
 
-CMD /vitalam-node/vitalam-node
+ENTRYPOINT [ "/vitalam-node/vitalam-node" ]
 
 EXPOSE 30333 9933 9944
 
-# docker run -it --rm -h node-0 -e IDENTITY=dev -e WS=true -p 30333:30333 -p 9944:9944 -p 9933:9933 vitalam-substrate-node ./run.sh
+#docker run -it --rm -h node-0 -e IDENTITY=dev -e WS=true -p 30333:30333 -p 9944:9944 -p 9933:9933 vitalam-substrate-node ./run.sh
```
mattdean-digicatapult, line 29: This comment is now spurious so can be removed.

mattdean-digicatapult (changes requested):
I tiny change and then I think we're good

mattdean-digicatapult:
@danielcooperxyz @dblane-digicatapult I've just looked and this will break the release workflow as unfortunately the docker build runs before the release build. Is this worth fixing in this PR?

mattdean-digicatapult (approved):

PR #20, opened 2022-01-12
Add TokenId to MetadataValue

jonmattgray:
Makes it possible for metadata to be of the existing `TokenId` type.

- [x] add `TokenId` to the MetadataValue enum in the runtime
- [x] add `TokenId` to the MetadataValue enum in the mock
- [x] add pallet test and change a couple of others to include examples of `TokenId` metadata
- [x] version bumps

### DIFF
--- a/pallets/simple-nft/src/mock.rs
+++ b/pallets/simple-nft/src/mock.rs
@@ -62,7 +62,7 @@ impl system::Config for Test {
 }
 
 parameter_types! {
-    pub const MaxMetadataCount: u32 = 3;
+    pub const MaxMetadataCount: u32 = 4;
 }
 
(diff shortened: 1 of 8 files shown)


pallets/simple-nft/src/mock.rs
```diff
@@ -62,7 +62,7 @@ impl system::Config for Test {
 }
 
 parameter_types! {
-    pub const MaxMetadataCount: u32 = 3;
+    pub const MaxMetadataCount: u32 = 4;
```
jonmattgray, line 65: increased this so the test for all metadata types (now 4) in one token still worked

mattdean-digicatapult (approved):
LGTM

PR #21, opened 2022-01-14
Use correct build target name from new image

danielcooperxyz:
The new docker build workflow needs a small fix for the new dockerfile, due to name changes.

Questions for possible changes that we can include to tidy up the workflow:
- Remove sscache on line 108?
- Remove buildkit cache on line 119?


mattdean-digicatapult (approved):
I think that will do it. Just get approval from @dblane-digicatapult and then this can merge

dblane-digicatapult (approved):

PR #22, opened 2022-01-17
Pre release workflow

danielcooperxyz:
Add pre-release versioning for feature branches

- [x] Install tomlq, via cargo, if not found 
- [x] Define pre-release version format e.g. `v2.6.1-feature/pre-release-workflow`
- [x] Specfiy a build as pre-release if `branch_name != main`
- [x] Trigger docker builds for pre release versions

### DIFF
--- a/.github/workflows/check-and-test.yml
+++ b/.github/workflows/check-and-test.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           fetch-depth: 0
       - run: git fetch --tags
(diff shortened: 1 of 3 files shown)


mattdean-digicatapult (changes requested):
Some changes from a first pass. We also need to look at cache keys here

mattdean-digicatapult (approved):

PR #24, opened 2022-01-19
Feature/pre release workflow

danielcooperxyz:
Fix issue with the old versions being overwritten


mattdean-digicatapult (commented):
I'm still not sure about the logic here. If `main` is on a pre-release version and we merge a PR not changing the version, no release should be overwritten. I believe as things stand this will happen. I think the actual change here may need to be to the workflow to basically never build if we don't have a new version and we're on `main`

mattdean-digicatapult (approved):

PR #27, opened 2022-01-26
Feature/helm chart

dblane-digicatapult:
Implementing helm chart

### DIFF
--- a/.github/workflows/release-helm.yml
+++ b/.github/workflows/release-helm.yml
@@ -111,7 +111,6 @@ jobs:
       is_new_version: ${{ steps.get_version.outputs.IS_NEW_VERSION }}
       version: ${{ steps.get_version.outputs.VERSION }}
       build_date: ${{ steps.get_version.outputs.BUILD_DATE }}
-      is_prerelease: ${{ steps.get_version.outputs.IS_PRERELEASE }}
     steps:
       - uses: actions/checkout@v2
       - run: git fetch --depth=1 --tags origin
--- a/.github/workflows/release-helm.yml
+++ b/.github/workflows/release-helm.yml
@@ -1,0 +1,147 @@
+name: Release helm charts
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - 'helm/vitalam-node'
+
+jobs:
+
+  preconditions:
+    runs-on: ubuntu-latest
+    outputs:
+      repo_name: ${{ steps.repo_ids.outputs.REPO_NAME }}
+      org_name: ${{ steps.repo_ids.outputs.ORG_NAME }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Check token
+        run: |
+          if [ -z "${{ secrets.GITHUB_TOKEN }}"]; then
+            echo "Must provide a GITHUB_TOKEN secret in order to run release workflow"
+            exit 1
+          fi
+      - name: Get repository identifiers
+        id: repo_ids
+        run: |
+          REPO_NAME=$(echo "${{ github.event.repository.name }}" | tr '[:upper:]' '[:lower:]')
+          ORG_NAME=$(echo "${{ github.event.repository.owner.name }}" | tr '[:upper:]' '[:lower:]')
+          echo "::set-output name=REPO_NAME::$REPO_NAME"
+          echo "::set-output name=ORG_NAME::$ORG_NAME"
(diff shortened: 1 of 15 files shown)


helm/vitalam-node/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: node
+description: A Helm chart to deploy Substrate/Polkadot nodes
+type: application
+version: 0.23.0
+appVersion: "0.0.1"
```
mattdean-digicatapult, line 6: I think we should probably update these. Note that the helm version is now decoupled from the app version so that should also be the case here right?

helm/vitalam-node/.helmignore
```diff
@@ -0,0 +1,2 @@
+/examples
+/scripts
```
mattdean-digicatapult, line 2: What does `.helmignore` do?

helm/vitalam-node/README.md
```diff
@@ -0,0 +1,103 @@
+# Substrate/Polkadot node helm chart
```
mattdean-digicatapult, line 1: We should probably be updating this a bit

.github/workflows/release-helm.yml
```diff
@@ -0,0 +1,147 @@
...
+        run: ct install  --charts helm/${{ needs.preconditions.outputs.repo_name }}/
+
+  check-version:
+    name: "Check version"
+    runs-on: ubuntu-latest
+    outputs:
+      is_new_version: ${{ steps.get_version.outputs.IS_NEW_VERSION }}
+      version: ${{ steps.get_version.outputs.VERSION }}
+      build_date: ${{ steps.get_version.outputs.BUILD_DATE }}
+      is_prerelease: ${{ steps.get_version.outputs.IS_PRERELEASE }}
```
mattdean-digicatapult, line 114: we're not using `is_prelease` so this should probably go

scripts/check-helm-version.sh
```diff
@@ -0,0 +1,46 @@
...
+  echo "##[set-output name=VERSION;]$CURRENT_VERSION"
+  echo "##[set-output name=BUILD_DATE;]$(date -u +'%Y-%m-%dT%H:%M:%SZ')"
+  echo "##[set-output name=IS_NEW_VERSION;]true"
+  if [[ $CURRENT_VERSION =~ [-] ]]; then
+    echo "##[set-output name=IS_PRERELEASE;]true"
+    echo "##[set-output name=NPM_RELEASE_TAG;]next"
+  else
+    echo "##[set-output name=IS_PRERELEASE;]false"
+    echo "##[set-output name=NPM_RELEASE_TAG;]latest"
+  fi
```
mattdean-digicatapult, line 43: `npm` is certainly irrelevant here as is the prerelease thing. Probably just delete this entirely?

scripts/check-helm-version.sh
```diff
@@ -0,0 +1,46 @@
+#! /bin/bash
+
+# exit when any command fails
+set -e
+
+function check_version_greater () {
+  local current=$1
+  local git_versions="${2:-0.0.0}"
```
mattdean-digicatapult, line 8: These aren't git versions so naming here is confusing. Perhaps `published_versions`?

mattdean-digicatapult (changes requested):

helm/vitalam-node/Chart.yaml
dblane-digicatapult: This is updated from the release that we forked from.
https://github.com/paritytech/helm-charts/blob/main/charts/node/Chart.yaml

helm/vitalam-node/.helmignore
dblane-digicatapult: Spurious file, deleted

helm/vitalam-node/Chart.yaml
mattdean-digicatapult: Hang on so the argument here is to not make this a `vitalam-node` chart but a generic substrate one?

dblane-digicatapult: I assumed given the chart is a fork we increment from the currently released versions, rather than start our own versioning.

mattdean-digicatapult: Alright seems fair

mattdean-digicatapult (approved):
Not perfect but I get the plan is that this should be temporary. Look forward to moving to the parity chart once fixed

PR #29, opened 2022-01-26
Bug/helm test lint

dblane-digicatapult:
Should fix the docker build in helm-release workflow
Fixes linting

### DIFF
--- a/.github/workflows/release-helm.yml
+++ b/.github/workflows/release-helm.yml
@@ -116,6 +116,8 @@ jobs:
       is_new_version: ${{ steps.get_version.outputs.IS_NEW_VERSION }}
       version: ${{ steps.get_version.outputs.VERSION }}
       build_date: ${{ steps.get_version.outputs.BUILD_DATE }}
+      toolchain: ${{ steps.get_toolchain.outputs.TOOLCHAIN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release-helm.yml
```diff
@@ -120,6 +125,10 @@ jobs:
         id: get_version
         run: ./scripts/check-helm-version.sh
         shell: bash
+      - name: Get Versions
+        run: |
+          echo "##[set-output name=TOOLCHAIN;]$(cat rust-toolchain)"
+          echo "##[set-output name=VERSION;]$(tomlq .package.version node/Cargo.toml | tr -d '"')"
```
mattdean-digicatapult, line 131: I think there's some confusion here. This version will currently never be used right because:

```
version: ${{ steps.get_version.outputs.VERSION }}
```

Do you need the `check-version-helm` script to output two versions `HELM_VERSION` and `APP_VERSION`?

mattdean-digicatapult (changes requested):

.github/workflows/release-helm.yml
dblane-digicatapult: Yes I just noticed this.

mattdean-digicatapult (approved):
This is pretty broke and we need to sort out these workflows. For now this is fine

PR #34, opened 2022-01-27
First stab at the process validation pallet

mattdean-digicatapult:
This PR lays the foundations for the process validation storage and evaluation pallet. There's no real functionality yet but what I have done is:

* created a new pallet `pallet-process-validation` which will store and manage the creation/disabling of process restrictions
* loosely coupled the new pallet to `pallet-simple-nft` via a shared trait and a struct that can describe a process input or output `ProcessIO`
* boiler plate for tests, benchmarking, weights

~I don't believe that this change will change any of the types used by the API~ Types still need updating

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -329,8 +329,20 @@ impl pallet_simple_nft::Config for Runtime {
     type RoleKey = Role;
     type TokenMetadataKey = [u8; 32];
     type TokenMetadataValue = MetadataValue<Self::TokenId>;
+    type ProcessValidator = ProcessValidation;
     type WeightInfo = pallet_simple_nft::weights::SubstrateWeight<Runtime>;
     type MaxMetadataCount = MaxMetadataCount;
+}
+
+impl pallet_process_validation::Config for Runtime {
+    type Event = Event;
+    type ProcessIdentifier = [u8; 32];
+    type CreateProcessOrigin = EnsureRoot<AccountId>;
+    type DisableProcessOrigin = EnsureRoot<AccountId>;
+    type WeightInfo = pallet_process_validation::weights::SubstrateWeight<Runtime>;
+    type RoleKey = Role;
+    type TokenMetadataKey = [u8; 32];
+    type TokenMetadataValue = MetadataValue<u128>;
 }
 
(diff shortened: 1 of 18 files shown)


nitro-marky (approved):
Looks like a great starting point (from my current Substrate knowledge!!).

danielcooperxyz (commented):
Looks like a lot of boilerplate to get an additional pallet in place.

It's good to see how the process validation is hooked up in the runtime.

It's a shame the weight info is duplicated like that, is it a substrate convention to name it that way or could we name them uniquely? i.e. `SimplePalletNFTWeightInfo`, `ProcessValidationWeightInfo`

jonmattgray:
Perhaps not worth it in this PR but it would be nice if the structure of `simple-nft` pallet matched the other pallets i.e. moving `tests.rs` into a folder (and splitting them up a bit?) and renaming `mock.rs` -> `tests.rs`

runtime/src/lib.rs
```diff
@@ -329,10 +327,22 @@ impl pallet_simple_nft::Config for Runtime {
...
 }
 
+impl pallet_process_validation::Config for Runtime {
+    type Event = Event;
+    type ProcessIdentifier = [u8; 32];
+    type CreateProcessOrigin = EnsureRoot<AccountId>;
+    type DisableProcessOrigin = EnsureRoot<AccountId>;
+    type WeightInfo = pallet_process_validation::weights::SubstrateWeight<Runtime>;
+    type RoleKey = Role;
+    type TokenMetadataKey = [u8; 32];
```
jonmattgray, line 342: could use a common type to match with `pallet_simple_nft`

`TokenMetadataValue` too

mattdean-digicatapult:
@danielcooperxyz 

> It's a shame the weight info is duplicated like that, is it a substrate convention to name it that way or could we name them uniquely? i.e. `SimplePalletNFTWeightInfo`, `ProcessValidationWeightInfo`

The point I guess is the weight info types are completely separate so the type checker will always catch any discrepency that might come up by accident. I've never seen a `WeightInfo` called anything other than a `WeightInfo` hence I just followed that pattern. You could call that a convension...

jonmattgray (approved):

PR #36, opened 2022-02-01
Feature/process validation usage

mattdean-digicatapult:
Optionally enforces process validation restrictions using the `process-validation` pallet. The changes in this PR are:

* implement `validate_process` logic in `pallet-process-validation`
* implement a `SenderOwnsAllInputs` restriction in `Restriction` so that restrictions can be tested properly
* Add in `process` parameter to `run_process` extrinsic in `pallet-simple-nft`. The optionally then uses the configured process validation logic instead of the pre-existing behaviour (which is persisted if `process == None`
* Test updates to cover the above

Todo:

- [x] update readme with current types
- [x] update readme to document usage of `process` argument in `pallet-simple-nft`
- [x] readme for `pallet-process-validation`

One other point for consideration is whether we should implement the current restriction set in the `Restriction` enum. This would involve implmenting:

* A restriction that all outputs must have a uniquely defined `original_id` or it should be `None`
* A restriction that the `Owner` role must be set on all outputs
* Possibly a restriction to enforce `original_id` behaviour

### DIFF
--- a/.github/workflows/check-and-test.yml
+++ b/.github/workflows/check-and-test.yml
@@ -101,7 +101,8 @@ jobs:
           RUSTC_WRAPPER: ${{ runner.temp }}/sccache
           SCCACHE_DIR: ${{ runner.temp }}/cache
         with:
-          command: fmt --check
+          command: fmt
+          args: --check
       - name: Cargo check
         uses: actions-rs/cargo@v1
         env:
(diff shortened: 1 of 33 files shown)


jonmattgray (approved):

darrylmorton-digicatapult (approved):

dblane-digicatapult (approved):
You should probably bump the `image.tag` in the helm chart values.yaml and bump the helm chart minor version.

danielcooperxyz (approved):

PR #37, opened 2022-02-04
Feature/in 121

n3op2:
### What

- added a new struct and a single storage map method with the process id as primary key - `Version`.
```rust
truct {
  process_id: [u8; 32] // process_id
  version: i32 // this should be changed to ProcessVersion type 
}
```
- some unit tests for the `create process` method and `disable_process`
- a couple of helper functions
- some logic for prevention of spamming events

### Test coverage
```
~/workspace/vitalam-node (feature/in-121) $ RUST_BACKTRACE=1 cargo watch -x 'test  -p pallet-process-validation -- --nocapture'
[Running 'cargo test  -p pallet-process-validation -- --nocapture']
   Compiling pallet-process-validation v1.0.0 (/Users/PMichelevicius/workspace/vitalam-node/pallets/process-validation)
    Finished test [unoptimized + debuginfo] target(s) in 4.33s
     Running unittests (target/debug/deps/pallet_process_validation-73f8bab8c750bef3)

running 12 tests
test tests::__construct_runtime_integrity_test::runtime_integrity_tests ... ok
test tests::disable_process::returns_error_if_origin_validation_fails_and_no_data_added ... ok
test tests::disable_process::returns_error_if_process_does_not_exist ... ok
test tests::create_process::updates_version_correctly_for_new_process_and_dispatches_event ... ok
test tests::create_process::updates_version_correctly_for_existing_proces_and_dispatches_event ... ok
test tests::create_process::handles_if_process_exists_for_the_new_version ... ok
test tests::create_process::for_existing_process_it_mutates_an_existing_version ... ok
test tests::disable_process::disables_process_and_dispatches_event ... ok
test tests::create_process::if_no_version_found_it_should_return_default_and_insert_new_one ... ok
test tests::disable_process::returns_error_if_process_is_already_disabled ... ok
test tests::create_process::returns_error_if_origin_validation_fails_and_no_data_added ... ok
test tests::create_process::sets_versions_correctly_for_multiple_processes ... ok

test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

   Doc-tests pallet-process-validation

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

[Finished running. Exit status: 0]
```

### TODO
- [x] - disable_process - will pair with Dan
- [x] - Errors and maybe some handling
- [x] - expand on unit tests coverage - create_process
- [x] - write unit tests for disable_process - Dan
- [x] - sort some types and a general clean up
- [x] - resolve conflicts

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -25,7 +25,7 @@ structopt = '0.3.8'
 hex-literal = "0.3.1"
 bs58 = "0.4.0"
 
-vitalam-node-runtime = { path = '../runtime', version = '2.2.0' }
+vitalam-node-runtime = { path = '../runtime', version = '2.2.1' }
 
 # Substrate dependencies
 frame-benchmarking = '3.0.0'
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -6,7 +6,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/vitalam-node/'
 name = 'vitalam-node'
-version = '2.8.1'
+version = '2.8.2'
 
 [[bin]]
 name = 'vitalam-node'
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -6,7 +6,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/vitalam-node/'
 name = 'vitalam-node'
-version = '2.7.1'
+version = '2.8.0'
 
 [[bin]]
 name = 'vitalam-node'
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -6,7 +6,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/vitalam-node/'
 name = 'vitalam-node'
-version = '2.8.0'
+version = '2.8.1'
 
 [[bin]]
 name = 'vitalam-node'
--- a/pallets/process-validation/src/lib.rs
+++ b/pallets/process-validation/src/lib.rs
@@ -1,9 +1,9 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use codec::{Decode, Encode};
-use frame_support::{dispatch::EncodeLike, Parameter};
+use frame_support::Parameter;
 pub use pallet::*;
-use sp_runtime::traits::AtLeast32Bit;
+use sp_runtime::traits::{AtLeast32Bit, One};
 use sp_std::prelude::*;
 
 use vitalam_pallet_traits::{ProcessIO, ProcessValidator};
@@ -30,14 +30,6 @@ impl Default for ProcessStatus {
         ProcessStatus::Disabled
     }
 }
-
-#[derive(Encode, Decode, Clone, PartialEq)]
-#[cfg_attr(feature = "std", derive(Debug))]
-struct Version {
-    version: i32,
-}
-
-impl EncodeLike<i32> for Version {}
 
 #[derive(Encode, Default, Decode, Clone, PartialEq)]
 #[cfg_attr(feature = "std", derive(Debug))]
@@ -66,7 +58,7 @@ pub mod pallet {
 
         // The primary identifier for a process (i.e. it's name, and version)
         type ProcessIdentifier: Parameter;
-        type ProcessVersion: Parameter + AtLeast32Bit;
+        type ProcessVersion: Parameter + AtLeast32Bit + Default;
 
         // Origins for calling these extrinsics. For now these are expected to be root
         type CreateProcessOrigin: EnsureOrigin<Self::Origin>;
@@ -90,20 +82,30 @@ pub mod pallet {
     /// Storage map definition
     #[pallet::storage]
     #[pallet::getter(fn process_model)] // not sure about name, store?, map?
-    pub(super) type ProcessModel<T: Config> =
-        StorageDoubleMap<_, Blake2_128Concat, T::ProcessIdentifier, Blake2_128Concat, i32, Process, ValueQuery>;
+    pub(super) type ProcessModel<T: Config> = StorageDoubleMap<
+        _,
+        Blake2_128Concat,
+        T::ProcessIdentifier,
+        Blake2_128Concat,
+        T::ProcessVersion,
+        Process,
+        ValueQuery,
+    >;
 
     #[pallet::storage]
     #[pallet::getter(fn version_model)]
-    pub(super) type VersionModel<T: Config> = StorageMap<_, Blake2_128Concat, T::ProcessIdentifier, i32, ValueQuery>;
+    pub(super) type VersionModel<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::ProcessIdentifier, T::ProcessVersion, ValueQuery>;
 
     #[pallet::event]
+    // looking by the type, same type for multiple things - bnad idea
+    // #[pallet::metadata(Vec<u8> = "a", Vec<u16> = "b")]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
         // id, version, restrictions, is_new
-        ProcessCreated(T::ProcessIdentifier, i32, Vec<Restriction>, bool),
-        //id, version, updated
-        ProcessDisabled(T::ProcessIdentifier, i32),
+        ProcessCreated(T::ProcessIdentifier, T::ProcessVersion, Vec<Restriction>, bool),
+        //id, version
+        ProcessDisabled(T::ProcessIdentifier, T::ProcessVersion),
     }
 
     #[pallet::error]
@@ -126,11 +128,15 @@ pub mod pallet {
             restrictions: Vec<Restriction>,
         ) -> DispatchResultWithPostInfo {
             T::CreateProcessOrigin::ensure_origin(origin)?;
-            let new_version: i32 = Pallet::<T>::_update_version(id.clone()).unwrap();
-            Pallet::<T>::_persist_process(&id, &new_version, restrictions.clone());
+            let new_version: T::ProcessVersion = Pallet::<T>::_update_version(id.clone()).unwrap();
+            Pallet::<T>::persist_process(&id, &new_version, restrictions.clone());
 
-            // wrap event data into a struct or some other data structure?
-            Self::deposit_event(Event::ProcessCreated(id, new_version, restrictions, new_version == 1));
+            Self::deposit_event(Event::ProcessCreated(
+                id,
+                new_version.clone(),
+                restrictions,
+                new_version == One::one(),
+            ));
 
             return Ok(().into());
         }
@@ -139,22 +145,20 @@ pub mod pallet {
         pub(super) fn disable_process(
             origin: OriginFor<T>,
             id: T::ProcessIdentifier,
-            version: i32,
+            version: T::ProcessVersion,
         ) -> DispatchResultWithPostInfo {
             T::DisableProcessOrigin::ensure_origin(origin)?;
 
             ensure!(
-                <ProcessModel<T>>::contains_key(&id, version),
+                <ProcessModel<T>>::contains_key(&id, version.clone()),
                 Error::<T>::NonExistingProcess,
             );
 
             // TODO check if there is any version for this process
-            ensure!(<VersionModel<T>>::contains_key(&id), Error::<T>::InvalidVersion,);
-
-            // Question for Matt, whether status is already disable should be an error or not
-            // also for the above ensure! macros
+            ensure!(<VersionModel<T>>::contains_key(&id), Error::<T>::InvalidVersion);
             Pallet::<T>::_disable_process(&id, &version)?;
 
+            // small and well defined (typed)
             Self::deposit_event(Event::ProcessDisabled(id, version));
             return Ok(().into());
         }
@@ -162,26 +166,29 @@ pub mod pallet {
 
     // helper methods
     impl<T: Config> Pallet<T> {
-        pub fn _get_version(id: T::ProcessIdentifier) -> i32 {
-            return <VersionModel<T>>::get(&id);
+        pub fn _get_version(id: &T::ProcessIdentifier) -> T::ProcessVersion {
+            let version: T::ProcessVersion = <VersionModel<T>>::get(&id);
+            return version;
         }
 
-        pub fn _update_version(id: T::ProcessIdentifier) -> Result<i32, ()> {
-            let version: i32 = Pallet::<T>::_get_version(id.clone());
-            let exists: bool = <ProcessModel<T>>::contains_key(id.clone(), version);
-            let new_version: i32 = match exists {
+        // rebase with master
+        pub fn _update_version(id: T::ProcessIdentifier) -> Result<T::ProcessVersion, ()> {
+            let version: T::ProcessVersion = Pallet::<T>::_get_version(&id);
+            let exists: bool = <ProcessModel<T>>::contains_key(&id, version.clone());
+            let new_version: T::ProcessVersion = match exists {
                 true => version,
-                false => version + 1,
+                false => version + One::one(),
             };
-            match new_version == 1 {
-                true => <VersionModel<T>>::insert(&id, Version { version: new_version }),
-                false => <VersionModel<T>>::mutate(&id, |v| *v = new_version),
+            match &new_version == &One::one() {
+                true => <VersionModel<T>>::insert(&id, new_version.clone()),
+                false => <VersionModel<T>>::mutate(&id, |v| *v = new_version.clone()),
             };
 
             return Ok(new_version);
         }
 
-        pub fn _persist_process(id: &T::ProcessIdentifier, version: &i32, restrictions: Restrictions) {
+        // remove underscores from helper methods
+        pub fn persist_process(id: &T::ProcessIdentifier, version: &T::ProcessVersion, restrictions: Restrictions) {
             <ProcessModel<T>>::insert(
                 id,
                 version,
@@ -192,7 +199,7 @@ pub mod pallet {
             );
         }
 
-        pub fn _disable_process(id: &T::ProcessIdentifier, version: &i32) -> Result<bool, Error<T>> {
+        pub fn _disable_process(id: &T::ProcessIdentifier, version: &T::ProcessVersion) -> Result<bool, Error<T>> {
             let process: Process = <ProcessModel<T>>::get(&id, &version);
             if process.status == ProcessStatus::Disabled {
                 return Err(Error::<T>::AlreadyDisabled);
--- a/pallets/process-validation/src/lib.rs
+++ b/pallets/process-validation/src/lib.rs
@@ -65,7 +65,6 @@ pub mod pallet {
     pub trait Config: frame_system::Config {
         /// The overarching event type.
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
-        type One: One;
         // The primary identifier for a process (i.e. it's name, and version)
         type ProcessIdentifier: Parameter;
         type ProcessVersion: Parameter + AtLeast32Bit + Default;
--- a/pallets/process-validation/src/lib.rs
+++ b/pallets/process-validation/src/lib.rs
@@ -23,6 +23,7 @@ use restrictions::Restriction;
 pub enum ProcessStatus {
     Disabled,
     Enabled,
+    None,
 }
 
 impl Default for ProcessStatus {
@@ -31,11 +32,20 @@ impl Default for ProcessStatus {
     }
 }
 
-#[derive(Encode, Default, Decode, Clone, PartialEq)]
+#[derive(Encode, Decode, Clone, PartialEq)]
 #[cfg_attr(feature = "std", derive(Debug))]
 pub struct Process {
     status: ProcessStatus,
     restrictions: Vec<Restriction>,
+}
+
+impl Default for Process {
+    fn default() -> Self {
+        Process {
+            status: ProcessStatus::None,
+            restrictions: vec![],
+        }
+    }
 }
 
 pub mod weights;
@@ -55,7 +65,7 @@ pub mod pallet {
     pub trait Config: frame_system::Config {
         /// The overarching event type.
         type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
-
+        type One: One;
         // The primary identifier for a process (i.e. it's name, and version)
         type ProcessIdentifier: Parameter;
         type ProcessVersion: Parameter + AtLeast32Bit + Default;
@@ -98,12 +108,7 @@ pub mod pallet {
         StorageMap<_, Blake2_128Concat, T::ProcessIdentifier, T::ProcessVersion, ValueQuery>;
 
     #[pallet::event]
-    // looking by the type, same type for multiple things - bnad idea
-    #[pallet::metadata(
-        ProcessIdentifier = "Process Id",
-        ProcessVersion = "Process Version",
-        bool = "Is New"
-    )]
+    #[pallet::metadata(ProcessIdentifier = "ProcessId", ProcessVersion = "ProcessVersion", bool = "IsNew")]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
         // id, version, restrictions, is_new
@@ -114,6 +119,8 @@ pub mod pallet {
 
     #[pallet::error]
     pub enum Error<T> {
+        // process already exists, investigate
+        AlreadyExists,
         // attempting to disable non-existing process
         NonExistingProcess,
         // process is already disabled
@@ -132,14 +139,14 @@ pub mod pallet {
             restrictions: Vec<Restriction>,
         ) -> DispatchResultWithPostInfo {
             T::CreateProcessOrigin::ensure_origin(origin)?;
-            let new_version: T::ProcessVersion = Pallet::<T>::update_version(id.clone()).unwrap();
-            Pallet::<T>::persist_process(&id, &new_version, restrictions.clone());
+            let version: T::ProcessVersion = Pallet::<T>::update_version(id.clone()).unwrap();
+            Pallet::<T>::persist_process(&id, &version, &restrictions)?;
 
             Self::deposit_event(Event::ProcessCreated(
                 id,
-                new_version.clone(),
+                version.clone(),
                 restrictions,
-                new_version == One::one(),
+                version == One::one(),
             ));
 
             return Ok(().into());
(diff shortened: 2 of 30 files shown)


pallets/process-validation/src/lib.rs
```diff
@@ -29,11 +30,29 @@ impl Default for ProcessStatus {
...
+#[derive(Encode, Decode, Clone, PartialEq)]
+#[cfg_attr(feature = "std", derive(Debug))]
+pub struct Version {
+    version: i32, // TODO: sort this type, should be included from trait
+}
+// TODO remove once type has been soprted <version>
+impl EncodeLike<i32> for Version {}
+
+
+#[derive(Encode, Decode, Clone, PartialEq)]
```
mattdean-digicatapult, line 42: You know this isn't need, but just a reminder that once the version is swapped out this can be removed

pallets/process-validation/src/lib.rs
```diff
@@ -29,11 +30,29 @@ impl Default for ProcessStatus {
...
+    restrictions: Vec<Restriction>,
+}
+
+impl Default for Process {
+    fn default() -> Process {
+        Process {
+            status: ProcessStatus::Disabled,
+            restrictions: vec![{ Restriction::None }],
+        }
+    }
```
mattdean-digicatapult, line 55: Why re-implment the `Default` trait? This was being derived by 

```
#[derive(Encode, Decode, Default, Clone, PartialEq)]
```

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
 
     /// Storage map definition
     #[pallet::storage]
-    #[pallet::getter(fn processes_by_id_and_version)]
-    pub(super) type ProcessesByIdAndVersion<T: Config> = StorageDoubleMap<
+    #[pallet::getter(fn get_process)]
+    pub(super) type GetProcess<T: Config> = StorageDoubleMap<
```
mattdean-digicatapult, line 104: I don't particularly like the storage name here as `GetProcess` feels like it describes a function for purely "getting". I think either `Processes` or as it was before `ProcessesByIdAndVersion` would feel closer

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
         _,
         Blake2_128Concat,
         T::ProcessIdentifier,
         Blake2_128Concat,
-        T::ProcessVersion,
+        i32,
```
mattdean-digicatapult, line 109: Reminder `T::ProcessVersion`

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
+        ValueQuery,
+    >;
+
+    #[pallet::storage]
+    #[pallet::getter(fn get_version)]
+    pub(super) type GetVersion<T: Config> = StorageMap<
+        _,
+        Blake2_128Concat,
+        T::ProcessIdentifier,
+        i32,
```
mattdean-digicatapult, line 120: `T::ProcessVersion`

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
     >;
 
     #[pallet::event]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
-        // TODO: implement correct events for extrinsics including params
-        ProcessCreated,
-        ProcessDisabled
+        // id, version, restrictions, is_new
+        ProcessCreated(T::ProcessIdentifier, i32, Vec<Restriction>, bool),
```
mattdean-digicatapult, line 128: `T::ProcessVersion`

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
     pub enum Error<T> {
         // TODO: implement errors for extrinsics
     }
 
+    // helper functions
+    fn get_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        return <GetVersion<T>>::get(&id);
+    }
+
+    fn uopdate_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
```
mattdean-digicatapult, line 142: typo on `update`. Can this also be a borrow to save on the `id.clone()` calls?

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
+    pub(super) type GetProcess<T: Config> = StorageDoubleMap<
         _,
         Blake2_128Concat,
         T::ProcessIdentifier,
         Blake2_128Concat,
-        T::ProcessVersion,
+        i32,
         Process,
+        ValueQuery,
+    >;
```
danielcooperxyz, line 112: can we remove the `ProcessVersion` type? or is that preferred over `i32`?

pallets/process-validation/src/lib.rs
danielcooperxyz, line 104: Thoughts behind this naming change?

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
+    // helper functions
+    fn get_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        return <GetVersion<T>>::get(&id);
+    }
+
+    fn uopdate_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        let version: i32 = get_version::<T>(id.clone());
+        let exists: Result<Process, ()> = <GetProcess<T>>::try_get(id.clone(),version);
+        let new_version: i32 = version + if exists.is_ok() { 0 } else { 1 };
+        <GetVersion<T>>::insert(
```
danielcooperxyz, line 146: GetVersion::insert feels like conflicting naming here

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
 
+    // helper functions
+    fn get_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        return <GetVersion<T>>::get(&id);
+    }
+
+    fn uopdate_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        let version: i32 = get_version::<T>(id.clone());
+        let exists: Result<Process, ()> = <GetProcess<T>>::try_get(id.clone(),version);
+        let new_version: i32 = version + if exists.is_ok() { 0 } else { 1 };
```
mattdean-digicatapult, line 145: Ok so here's the tricky bit when we use `T::ProcessVersion`. Firstly I would switch to `match` syntax as you won't then have to do any weird add zero trickery. Second we can refer to `1` by using `One::one()` which can be imported using `use sp_runtime::traits::One;`. All in all you'll end up with something like:

```rust
let new_version: T::ProcessVersion = match exists {
  Ok(version) => version + One::one(),
  Err(_) => Default::default()
};
```

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
     }
 
+    // helper functions
+    fn get_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        return <GetVersion<T>>::get(&id);
+    }
+
+    fn uopdate_version<T: Config>(id: T::ProcessIdentifier) -> i32 {
+        let version: i32 = get_version::<T>(id.clone());
+        let exists: Result<Process, ()> = <GetProcess<T>>::try_get(id.clone(),version);
```
mattdean-digicatapult, line 144: This check here I'm uncertain about. If this check fails we have a data integrity problem. You could probably just skip this (as it doens't help us see or solve the integrity problem anyway)

pallets/process-validation/src/lib.rs
```diff
@@ -79,50 +100,113 @@ pub mod pallet {
...
+        <GetVersion<T>>::insert(
+            &id,
+            Version {
+                version: new_version,
+            }
+        );
+
+        return new_version;
+    } 
+    fn persist_process<T: Config>(id: &T::ProcessIdentifier, version: &i32, restrictions: Restrictions) {
```
mattdean-digicatapult, line 155: `T::ProcessVersion`

pallets/process-validation/src/lib.rs
danielcooperxyz, line 142: typo here in `uopdate`?

mattdean-digicatapult (changes requested):
Alright this probably looks like a wall of changes but most of it is arround the switch to `T::ProcessVersion`. This is actually really close on the implmentation side, so well done! Couple of additional points:

1. Please rebase or merge in main. There's currently some churn from the `cargo fmt`
2. Tests need writing still, but I think you know that. Just a reminder

pallets/process-validation/src/lib.rs
n3op2: 1009%

pallets/process-validation/src/lib.rs
n3op2: No, this is something we need to implement. The idea is to have either a global trait in terms of process_validation scope

pallets/process-validation/src/lib.rs
n3op2: agreed, should rename to just a version and process maybe to keep it simple. Because this allows us to perform a various calls to the storage

pallets/process-validation/src/lib.rs
n3op2: Yes, so this is like a very draft that is not really there, just wanted share something with out and get some input

Replied to Dan's comment above in regards to the version type

pallets/process-validation/src/lib.rs
n3op2: mmm.... might need to read a bit more on Default trait and what it does, noted!

pallets/process-validation/src/lib.rs
n3op2: Yes! well spotted, just wanted to keep something simple, easier to remember, will update to Process and Version I guess because i see it as ORM/db model. Please let me know what would you prefer

pallets/process-validation/src/lib.rs
n3op2: So yestyerday I've tried but then it was just getting late and left it for today. Already got an idea how to implement it, even a few options

- add a trait to version
- make trait global to process validation
- pass it along

pallets/process-validation/src/lib.rs
n3op2: nice, thanks!

pallets/process-validation/src/lib.rs
n3op2: Was thinking that this could be handled by the pallet::error? Still to be implemented, plan for today, there is a little toido in the descriptionj

pallets/process-validation/src/lib.rs
n3op2: yes sir!

pallets/process-validation/src/lib.rs
mattdean-digicatapult: The idea @danielcooperxyz behind having it as a trait is to force consistency of types between versions referenced in the `pallet-simple-nft` `run_procss` extrinsic and the type used in this pallet

pallets/process-validation/src/lib.rs
mattdean-digicatapult: As discussed in slack the question really is what is the error condition? I don't think the previous version not existing is a huge problem and shouldn't error. What might be worth an error though is if the new proposed version already exists. That would be really bad. Not sure how we'll test that behaviour though 🤔

pallets/process-validation/src/lib.rs
n3op2: unfortunately, it can not, it looks like storage calls do drop the value from memory

pallets/process-validation/src/lib.rs
danielcooperxyz: wut

pallets/process-validation/src/tests/create_process.rs
```diff
@@ -1,10 +1,142 @@
...
-fn create_process_simple() {
+fn returns_error_if_origin_validation_fails_and_no_data_added() {
     new_test_ext().execute_with(|| {
-        assert_ok!(ProcessValidation::create_process(Origin::root()));
+        // update to check for error type
+        assert!(ProcessValidation::create_process(
+            Origin::none(),
+            PROCESS_ID1,
+            vec![{ None }]
+        ).is_err());
```
mattdean-digicatapult, line 24: I would recomment for this:

```rust
assert_noop!(
  ProcessValidation::create_process(Origin::none(), PROCESS_ID1, vec![{ None }],
  DispatchError::BadOrigin
);
```

As used in the `symmetric_key` pallet. That'll also assert that nothing has changed in storage I believe

pallets/process-validation/src/tests/create_process.rs
```diff
@@ -1,10 +1,142 @@
...
+fn if_no_version_found_it_should_return_default_and_insert_new_one() {
+    new_test_ext().execute_with(|| {
+        System::set_block_number(1);
+        assert_eq!(<VersionModel<Test>>::get(PROCESS_ID1), 0);
+        assert_ok!(ProcessValidation::create_process(
+            Origin::root(),
+            PROCESS_ID1,
+            vec![{ None }],
+        ));
+        assert_eq!(<VersionModel<Test>>::get(PROCESS_ID1), 1);
```
mattdean-digicatapult, line 44: You're only asserting against the `VersionModel` and not the `ProcessModel`. I'd probably make sure the actual process assertion happened correctly as well

pallets/process-validation/src/tests/create_process.rs
```diff
@@ -1,10 +1,142 @@
...
+        assert_eq!(System::events()[0].event, expected);
+    });
+}
+
+#[test]
+fn creates_new_process_and_dispatches_event() {
+    new_test_ext().execute_with(|| {
+        System::set_block_number(1);
+    });
+}
```
mattdean-digicatapult, line 142: Remove this as it's now redundent

pallets/process-validation/src/lib.rs
```diff
@@ -79,53 +90,110 @@ pub mod pallet {
...
+                id,
+                version,
+                Process {
+                    restrictions: restrictions.clone(),
+                    ..Default::default()
+                },
+            );
+        }
+     }
 }
```
mattdean-digicatapult, line 194: I like separating this out a lot! One minor comment is on the accessibility and naming. Do these need to be `pub fn` and I'd suggest dropping the leading underscore. In rust leading underscrores are used to make unused variables so I can see this being confusing

pallets/process-validation/src/lib.rs
```diff
@@ -79,53 +90,110 @@ pub mod pallet {
...
+            // TODO implement errors
+            // TODO if no process for this version return latest along with the error message
+            if !<ProcessModel<T>>::contains_key(id.clone(), version) {
+                Self::deposit_event(Event::ProcessDisabled(id, version, false));
+                return Ok(().into());
+            }
+
+            let mut updated = false;
+
+            <ProcessModel<T>>::mutate(id.clone(), version, |process| {
```
mattdean-digicatapult, line 153: In answer to your slack question definitely mutate over get and insert. It's just more idiomatic given what we're doing

pallets/process-validation/src/lib.rs
```diff
@@ -79,53 +90,110 @@ pub mod pallet {
...
+            id: T::ProcessIdentifier,
+            restrictions: Vec<Restriction>,
+        ) -> DispatchResultWithPostInfo {
             T::CreateProcessOrigin::ensure_origin(origin)?;
+            let new_version: i32 = Pallet::<T>::_update_version(id.clone()).unwrap();
+            Pallet::<T>::_persist_process(&id, &new_version, restrictions.clone());
 
-            Self::deposit_event(Event::ProcessCreated);
-            Ok(().into())
+            // wrap event data into a struct?
```
mattdean-digicatapult, line 130: Could do but probably the most idiomatic way of assigning meaning to the values in an event is with event metadata as used in the symmetric key pallet:

```rust
#[pallet::event]
#[pallet::metadata(Vec<u8> = "Key")]
#[pallet::generate_deposit(pub(super) fn deposit_event)]
pub enum Event<T: Config> {
    // key was updated.
    UpdateKey(Vec<u8>),
}
```

mattdean-digicatapult (commented):
Tests are looking good. Many points from the first review still remain but we're headed in the right direction. Few comments to look at

pallets/process-validation/src/lib.rs
n3op2: Renamed to ProcessModel and VersionModel

pallets/process-validation/src/lib.rs
n3op2: did in a slightly different way as `Ok(version)` - was expecting a Process struct

pallets/process-validation/src/lib.rs
n3op2: Replaced with `contains_key` which returns a boolean value. I guess it makes more sense, pushing shortly!

pallets/process-validation/src/tests/create_process.rs
n3op2: Nice, event better, just added some errors, updating

n3op2: refactored, pushing shortly

pallets/process-validation/src/tests/create_process.rs
n3op2: Updated, also checking `System:events()` The reason was that storage is mocked already and create_process has returned 'Ok()' so querying storage should give correct data, well at least this is what I think.

pallets/process-validation/src/lib.rs
n3op2: They do for couple of reasons 
1. allows to use iin pallet otherwise you cannot call `Pallet::_helper_fn` and allows to use in unit testing, thought private should cover but it wasn't will give it another go shortly

pallets/process-validation/src/lib.rs
n3op2: :+1: this has been updated a little.

pallets/process-validation/src/lib.rs
n3op2: This is what I wanted!

pallets/process-validation/src/lib.rs
```diff
@@ -79,49 +89,120 @@ pub mod pallet {
...
+    impl<T: Config> Pallet<T> {
+        pub fn _get_version(id: T::ProcessIdentifier) -> i32 {
+            return <VersionModel<T>>::get(&id);
+        }
+
+        pub fn _update_version(id: T::ProcessIdentifier) -> Result<i32, ()> {
+            let version: i32 = Pallet::<T>::_get_version(id.clone());
+            let exists: bool = <ProcessModel<T>>::contains_key(id.clone(), version);
+            let new_version: i32 = match exists {
+                true => version,
```
jonmattgray, line 173: sorry if I'm misunderstanding the code but I'd expect these to be the other way round, if the process already exists version number increases?

pallets/process-validation/src/lib.rs
```diff
@@ -79,49 +89,120 @@ pub mod pallet {
...
 
     #[pallet::event]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
-        // TODO: implement correct events for extrinsics including params
-        ProcessCreated,
-        ProcessDisabled,
+        // id, version, restrictions, is_new
+        ProcessCreated(T::ProcessIdentifier, i32, Vec<Restriction>, bool),
+        //id, version, updated
```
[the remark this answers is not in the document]
n3op2: It has been removed:) it was just temp Dan's implementation. Please ignore

pallets/process-validation/src/lib.rs
n3op2: @jonmattgray I know, this is something I would need some clarity, looks like it works other way around not sure why tho. Need to do some more invstigation tho

pallets/process-validation/src/lib.rs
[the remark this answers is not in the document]
n3op2: Going to implement as per Matt's suggestion #[pallet::metadata] for the events

pallets/process-validation/src/lib.rs
```diff
@@ -79,8 +81,8 @@ pub mod pallet {
 
     /// Storage map definition
     #[pallet::storage]
-    #[pallet::getter(fn processes_by_id_and_version)]
-    pub(super) type ProcessesByIdAndVersion<T: Config> = StorageDoubleMap<
+    #[pallet::getter(fn process_model)] // not sure about name, store?, map?
```
mattdean-digicatapult, line 84: Remove comment

danielcooperxyz (commented):
You've done really well to pick this up and make something meaningful. Well done!

I don't have much to comment regarding the logic and types/structs. I noticed there are a lot of comments, some seem anecdotal whereas some seem more formal, also the formal comments are not always consistently applied i.e. method and variable documentation.

Is there a standard for code comment documentation in rust/substrate?

pallets/process-validation/src/lib.rs
```diff
@@ -90,38 +92,140 @@ pub mod pallet {
...
+    #[pallet::getter(fn version_model)]
+    pub(super) type VersionModel<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::ProcessIdentifier, T::ProcessVersion, ValueQuery>;
+
     #[pallet::event]
+    // looking by the type, same type for multiple things - bnad idea
+    #[pallet::metadata(
+        ProcessIdentifier = "Process Id",
+        ProcessVersion = "Process Version",
+        bool = "Is New"
```
mattdean-digicatapult, line 105: Not sure if these can/should include spaces. You may want to look into metadata usages elsewhere in substrate to get an idea of what's idiomatic

pallets/process-validation/src/lib.rs
```diff
@@ -29,15 +31,14 @@ impl Default for ProcessStatus {
     }
 }
 
-#[derive(Encode, Decode, Default, Clone, PartialEq)]
+#[derive(Encode, Default, Decode, Clone, PartialEq)]
```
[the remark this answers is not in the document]
n3op2: if `cargo fmt` did not care much I don't either, it does not matter in which order, it's just an import* of traits for the module/struct etc

pallets/process-validation/src/lib.rs
```diff
@@ -90,38 +92,140 @@ pub mod pallet {
...
+        pub fn get_version(id: &T::ProcessIdentifier) -> T::ProcessVersion {
+            let version: T::ProcessVersion = <VersionModel<T>>::get(&id);
+            return version;
+        }
+
+        // rebase with master
+        pub fn update_version(id: T::ProcessIdentifier) -> Result<T::ProcessVersion, ()> {
+            let version: T::ProcessVersion = Pallet::<T>::get_version(&id);
+            let exists: bool = <ProcessModel<T>>::contains_key(&id, version.clone());
+            let new_version: T::ProcessVersion = match exists {
```
mattdean-digicatapult, line 175: The logic here is dodgey. I don't think `exists` is ever `true`. If `id` doesn't exist `version` will be the default value of `T::ProcessVersion` which for our impl is `0`. That will make `version` `1` which is what we say is new. This should definitely be cleaned up

mattdean-digicatapult (changes requested):

pallets/process-validation/src/lib.rs
n3op2: updated

pallets/process-validation/src/lib.rs
n3op2: yes sir!

pallets/process-validation/src/lib.rs
n3op2: Thanks,. going to back this up with the new test scenario

n3op2: This has been updated, going to mark as resolved

pallets/process-validation/src/lib.rs
```diff
@@ -90,38 +101,141 @@ pub mod pallet {
...
+            };
+
+            return Ok(version);
+        }
+
+        pub fn persist_process(
+            id: &T::ProcessIdentifier,
+            v: &T::ProcessVersion,
+            r: &Restrictions,
+        ) -> Result<bool, Error<T>> {
```
mattdean-digicatapult, line 196: Not going to fail the PR for this but as you only ever return `Some(true)` or `Err(...)` this would likely be better represented as `Result<(), Error<T>>`

mattdean-digicatapult (changes requested):
Really close couple of minor changes. Also need to consider the following:
* bump pallet version
* bump runtime version (including the chain runtime version to match in the runtime lib.rs)
* bump node version

pallets/process-validation/src/lib.rs
n3op2: it was a bug, updated

runtime/Cargo.toml
```diff
@@ -4,7 +4,7 @@ edition = '2018'
 license = 'Apache-2.0'
```
mattdean-digicatapult, line 4: upate dependency version for pallet-process-validation to `1.0.1`

node/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = '2018'
 license = 'Apache-2.0'
```
mattdean-digicatapult, line 6: bump vitalam-node-runtime version to `2.2.1`

pallets/process-validation/src/lib.rs
```diff
@@ -90,38 +101,142 @@ pub mod pallet {
...
+                    });
+                    return Ok(true);
+                }
+            };
+        }
+
+        pub fn validate_version_and_process(
+            id: &T::ProcessIdentifier,
+            version: &T::ProcessVersion,
+        ) -> Result<bool, Error<T>> {
```
mattdean-digicatapult, line 230: Same again

pallets/process-validation/src/lib.rs
```diff
@@ -90,38 +101,142 @@ pub mod pallet {
...
+                            restrictions: r.clone(),
+                            ..Default::default()
+                        },
+                    );
+                    return Ok(());
+                }
+            };
+        }
+
+        pub fn set_disabled(id: &T::ProcessIdentifier, version: &T::ProcessVersion) -> Result<bool, Error<T>> {
```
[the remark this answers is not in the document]
n3op2: 🤦  sorry been rushing a little i guess and did not see that this repeats

jonmattgray:
Sorry to be the README troll, but this pallet has reached a point where it should have a section in the README? Like the sections for other pallets detailing their storage + extrinsics

n3op2:
@jonmattgray - was actually thinking about this earlier in the morning, will add a README file

mattdean-digicatapult (approved):
@n3op2 I think this is good to merge. Don't worry about the readme, when I jump back to finishing off #36 I'll update the readme there. Good stuff

PR #38, opened 2022-02-07
Bug/Use original_id of parent for new token original_id

jonmattgray:
When multiple tokens are about the same thing e.g a PO, we want them all to have the same `original_id`. This PR changes the pallet to correctly assign `original_id` rather than the `id` of a parent to a new token's `original_id`.

e.g. assuming all these tokens are about the same thing, such as a `Quote`. This PR goes from this:
Token 1: `id: 1` `original_id: 1`
Token 2: `id: 2` `original_id: 1`
Token 3: `id: 3` `original_id: 2` // this is wrong

to this
Token 1: `id: 1` `original_id: 1`
Token 2: `id: 2` `original_id: 1`
Token 3: `id: 3` `original_id: 1` // this is correct

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3905,7 +3905,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-simple-nft"
-version = "2.1.1"
+version = "2.1.2"
 dependencies = [
  "frame-benchmarking",
  "frame-support",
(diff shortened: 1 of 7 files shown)


mattdean-digicatapult (changes requested):
Good catch. Note this introduces extra storage gets so weights should be updated

mattdean-digicatapult (approved):

PR #41, opened 2022-02-10
Inteli stage genesis file

mattdean-digicatapult:
Accompanies https://github.com/digicatapult/vitalam-flux-infra/pull/6


dblane-digicatapult (approved):

n3op2 (approved):

node/src/command.rs
```diff
@@ -49,7 +49,8 @@ impl SubstrateCli for Cli {
     fn load_spec(&self, id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
         Ok(match id {
             "dev" => Box::new(chain_spec::development_config()?),
-            "" | "vitalam" => Box::new(chain_spec::vitalam_config()?),
+            "" | "inteli-stage" => Box::new(chain_spec::inteli_stage_config()?),
+            "vitalam" => Box::new(chain_spec::vitalam_config()?),
```
danielcooperxyz, line 53: so our network runs all the chsins at once? or specify a single one?

mattdean-digicatapult: specify one, that's in a `match` statement

danielcooperxyz (commented):
why do we store genesis in the node repo and not the infra repo? is it compiled as part of the node binary?

mattdean-digicatapult:
> why do we store genesis in the node repo and not the infra repo? is it compiled as part of the node binary?

yep, it get's compiled in. That way we can just refer to the chain by name

PR #43, opened 2022-02-11
Adds types LoadBalancer and ClusterIP to valid service types

dblane-digicatapult:
Adds ability to use different service types `ClusterIP` and `LoadBalancer`
Logic to determine if to use a public addr
Fixes some bugs in the parity chart with regards to bash flow logic on Parachain section
Parachain functionality needs to be tested against `Polkadot` container in the coming week.

### DIFF
--- a/helm/vitalam-node/templates/service.yaml
+++ b/helm/vitalam-node/templates/service.yaml
@@ -23,7 +23,7 @@ spec:
       name: websocket-rpc
 ---
 {{range $i := until (.Values.node.replicas | int) }}
-{{- if $.Values.node.perNodeServices.createClusterIPService }}
+{{- if $.Values.node.perNodeServices.createApiService }}
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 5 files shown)


helm/vitalam-node/templates/service.yaml
```diff
@@ -46,36 +45,39 @@ spec:
...
-  externalTrafficPolicy: Local
+  {{ include "vitalam-node.serviceType" $ | nindent 2 }}
   selector:
     {{- $selectorLabels | nindent 4 }}
     statefulset.kubernetes.io/pod-name: {{ $fullname }}-{{ $i }}
   ports:
     - name: p2p
       port: 30333
       targetPort: 30333
+  {{- if and $.Values.node.collator.isParachain (or (eq $.Values.node.perNodeServices.p2pServiceType "LoadBalancer") (eq $.Values.node.perNodeServices.p2pServiceType "ClusterIP")) }}
```
danielcooperxyz, line 62: this is easy to read and understand 🤨 

are variables possible here?

dblane-digicatapult: Variables - no
Nested If statements but that is even uglier.

Helm does allow this template to be split over multiple lines however the linter blows up.  I'll need to investigate this at a later date.

mattdean-digicatapult (approved):
This is extremely ugly. Please create a backlog ticket to clean up these scripts so they are more readable

PR #49, opened 2022-02-23
IN-99 Restrict process inputs to a fixed number

nitro-marky:
This allows restriction on the number of inputs to be a fixed integral value.


pallets/process-validation/src/restrictions.rs
```diff
@@ -157,4 +159,114 @@ mod tests {
...
+            },
+        ];
+        let result = validate_restriction::<u64, u32, u32, u64>(
+            &Restriction::FixedNumberOfInputs { num_inputs: 3 },
+            &1u64,
+            &inputs,
+            &Vec::new(),
+        );
+        assert!(!result);
+    }
```
dblane-digicatapult, line 271: Move to `tests` folder?

dblane-digicatapult (changes requested):

pallets/process-validation/src/restrictions.rs
nitro-marky: Do you want me to move all tests or just mine?

jonmattgray:
The `Restriction` enum in the huge types definition in the README should be updated to include the new value. The types get used by Polkadot e.g. https://github.com/digicatapult/vitalam-api/blob/main/app/util/substrateApi.js

Until [this ticket](https://digicatapult.atlassian.net/browse/IN-111) is done, the types in the README of this repo are copy-pasted into other repos when we bump their `vitalam-node` versions

pallets/process-validation/src/restrictions.rs
mattdean-digicatapult: I wouldn't move the tests for now (perhaps later). I would however look at renaming the tests to include the restriction type. Your first test might be something like `fixed_number_of_inputs_restriction_matches_fixed_input_total`. I know it's verbose but it correctly describes the test case in a more global context

nitro-marky: I did move the tests but I'll undo

nitro-marky:
I've added `FixedNumberOfInputs` to the Readme Types @jonmattgray

README.md
```diff
@@ -157,7 +157,7 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
     "_enum": ["Disabled", "Enabled"]
   },
   "Restriction": {
-    "_enum": ["None", "SenderOwnsAllInputs"]
+    "_enum": ["None", "SenderOwnsAllInputs", "FixedNumberOfInputs"]
```
jonmattgray, line 160: might need a <u32> on the end? I'm not 100% sure

mattdean-digicatapult (changes requested):
A few high level todos:

- [ ] Update README with correct types changes. See https://polkadot.js.org/docs/api/start/types.extend#user-defined-enum for an example. Basically the enum is a struct which has a property `num_inputs` so you'll need to define a type for the struct as well
- [x] Update package version for `process-validation`
- [x] Update referenced package version for the `process-validation` crate in the `runtime` `Cargo.toml`
- [x] Update the `runtime` version in both the `runtime` `Cargo.toml` and the `spec_version` in the `runtime` `lib.rs`
- [x] Update the referenced runtime verison in the `node` and the `node` version

All version bumps should be minor version bumps and zero out the bug fix number

README.md
nitro-marky: I've updated the readme from what I've seen in the docs so should hopefully be right

README.md
```diff
@@ -157,7 +157,11 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
     "_enum": ["Disabled", "Enabled"]
   },
   "Restriction": {
-    "_enum": ["None", "SenderOwnsAllInputs"]
+    "_enum": {
+      "None": null,
+      "SenderOwnsAllInputs": null,
+      "FixedNumberOfInputs": "num_inputs<u32>"
```
mattdean-digicatapult, line 163: I may be wrong but I think you'll need a new top level type. For example it could read:

```json
  "Restriction": {
    "_enum": ["None", "SenderOwnsAllInputs"]
    "_enum": {
      "None": null,
      "SenderOwnsAllInputs": null,
      "FixedNumberOfInputs": "FixedNumberOfInputsRestriction"
  },
  "FixedNumberOfInputsRestriction": {
    "num_inputs": "u32"
  }
```

I may be wrong about that though (docs arent overly clear). The way to test is to go to https://polkadot.js.org/apps/  configure it to your node and put in the types under the develop tab. We can look at this tomorrow if you like and work out the rigth answer

danielcooperxyz:
> 

These should all be automated checks, no?

mattdean-digicatapult:
> These should all be automated checks, no?

@danielcooperxyz that would be an excellent idea yes. I wouldn't include that in this ticket though. It's also somewhat related to us standardising our release processes (IN-89 in Jira)

node/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/vitalam-node/'
 name = 'vitalam-node'
-version = '2.9.3'
+version = '3.0.0'
```
dblane-digicatapult, line 9: Why are we major version bumping the node?

nitro-marky: My bad, I'll change it to 10

pallets/process-validation/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/vitalam-node/'
 name = 'pallet-process-validation'
-version = "1.1.1"
+version = "1.2.1"
```
dblane-digicatapult, line 8: 1.2.0

runtime/Cargo.toml
```diff
@@ -4,7 +4,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/vitalam-node/'
 name = 'vitalam-node-runtime'
-version = '2.3.1'
+version = '2.4.1'
```
dblane-digicatapult, line 7: 2.4.0

runtime/Cargo.toml
```diff
@@ -26,7 +26,7 @@ serde = { features = ['derive'], optional = true, version = '1.0.101' }
 
 # local dependencies
 pallet-simple-nft = { version = '2.2.0', default-features = false, package = 'pallet-simple-nft', path = '../pallets/simple-nft' }
-pallet-process-validation = { version = '1.1.1', default-features = false, package = 'pallet-process-validation', path = '../pallets/process-validation' }
+pallet-process-validation = { version = '1.2.1', default-features = false, package = 'pallet-process-validation', path = '../pallets/process-validation' }
```
dblane-digicatapult, line 29: 1.2.0

runtime/src/lib.rs
```diff
@@ -93,7 +93,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("vitalam-node"),
     impl_name: create_runtime_str!("vitalam-node"),
     authoring_version: 1,
-    spec_version: 231,
+    spec_version: 241,
```
dblane-digicatapult, line 96: 240

node/Cargo.toml
dblane-digicatapult: 2.10.0

dblane-digicatapult (changes requested):

dblane-digicatapult (approved):

mattdean-digicatapult (approved):

PR #50, opened 2022-02-24
Feature/npm api package

darrylmorton-digicatapult:
- [x] npm module
- [x] api init/creation function
- [x] exports: api, enums, keyring
- [x] git workflows lint, depcheck, check-version, publish
- [x] readme
- [x] depcheck
- [x] version script in parent directory
- [x] version bump

Tested as a local npm module in the `vitalam-api` repo also with the `@polkadot/api` package removed.


mattdean-digicatapult (commented):
A few high level points:
* there shouldn't need to be further open-sourcing stuff as it's already top level
* I don't like libraries instantiating their own logger and inspecting environment variables. These should be passable as arguments to a function that wraps the API creation process.

Hopefully that makes sense

dblane-digicatapult (commented):
A few thoughts that we need to address (either now or later)

How do we test that this package accurately reflects the types the substrate API uses?  I.E a workflow that brings up the node and then runs the API light module against it?

Releasing (I know we discussed this) - Currently there is no way for our nodeJS applications to consume this library/module.  How are we intending to publish this so that they can?

vitalam-node-api/src/env.js
```diff
@@ -0,0 +1,19 @@
...
+const dotenv = require('dotenv')
+
+if (process.env.NODE_ENV === 'test') {
+  dotenv.config({ path: 'test/test.env' })
+} else {
+  dotenv.config({ path: '.env' })
+}
+
+const vars = envalid.cleanEnv(process.env, {
+  LEGACY_METADATA_KEY: envalid.str({ default: '' }),
```
jonmattgray, line 11: I think you can remove this env (and in README)? It's not used in the types definition anymore and will soon be completely gone from API with my current ticket to remove legacy code

scripts/check-vitalam-node-version.sh
```diff
@@ -6,20 +6,11 @@ set -e
...
-  local HELM_VALUES_TAG_VERSION=$(yq eval '.image.tag' ./helm/vitalam-api/values.yaml)
-  local HELM_CHART_VERSION=$(yq eval '.version' ./helm/vitalam-api/Chart.yaml)
-  local HELM_CHART_APP_VERSION=$(yq eval '.appVersion' ./helm/vitalam-api/Chart.yaml)
-  
-  if [ "$PACKAGE_VERSION" != "$PACKAGE_LOCK_VERSION" ] ||
-     [ "v$PACKAGE_VERSION" != "$HELM_VALUES_TAG_VERSION" ] ||
-     [ "$PACKAGE_VERSION" != "$HELM_CHART_VERSION" ] ||
-     [ "$PACKAGE_VERSION" != "$HELM_CHART_APP_VERSION" ]; then
+
+  if [ "$PACKAGE_VERSION" != "$PACKAGE_LOCK_VERSION" ]; then
```
dblane-digicatapult, line 10: You need to ensure this is the same value as the runtimes cargo.toml

.github/workflows/release-vitalam-node.yml
```diff
@@ -0,0 +1,141 @@
...
+    if: ${{ needs.check-version.outputs.is_new_version == 'true' }}
+      # Build github release
+      - name: Build release version
+        uses: softprops/action-gh-release@v1
+        with:
+          token: '${{ secrets.GITHUB_TOKEN }}'
+          tag_name: ${{ needs.check-version.outputs.version }}
+          prerelease: false
+          name: ${{ needs.check-version.outputs.version }}
+          generate_release_notes: true
```
dblane-digicatapult, line 121: You aren't actually releasing anything.

You need to do an `npm ci`, login to github package registry then do an `npm publish` or something to that effect

dblane-digicatapult (changes requested):

.github/workflows/release-vitalam-node.yml
dblane-digicatapult: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-github-packages

dscp-node/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@digicatapult/dscp-node",
+  "version": "3.0.1",
+  "description": "VITALam Node API types",
```
dblane-digicatapult, line 4: VitalAM Node

jonmattgray: thanks, thought I'd purged all the vitalams

PR #51, opened 2022-02-24
IN-100 Restrict number of outputs for a process

nitro-marky:
This allows restrictions on the number of outputs to a fixed integral value. Tested on Polkadot and all seems well

Screenshot 2022-02-24 at 14 03 45

### DIFF
--- a/pallets/process-validation/src/restrictions.rs
+++ b/pallets/process-validation/src/restrictions.rs
@@ -267,7 +267,7 @@ mod tests {
             &Restriction::FixedNumberOfOutputs { num_outputs: 1 },
             &1u64,
             &Vec::new(),
-            &_outputs, 
+            &_outputs,
         );
         assert!(!result);
     }
--- a/pallets/process-validation/src/restrictions.rs
+++ b/pallets/process-validation/src/restrictions.rs
@@ -12,6 +12,7 @@ pub enum Restriction {
     None,
     SenderOwnsAllInputs,
     FixedNumberOfInputs { num_inputs: u32 },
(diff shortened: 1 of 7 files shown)


pallets/process-validation/src/restrictions.rs
```diff
@@ -35,6 +36,7 @@ where
     match *restriction {
         Restriction::None => true, // TODO implement some actual restrictions
         Restriction::FixedNumberOfInputs { num_inputs } => return inputs.len() == num_inputs as usize,
+        Restriction::FixedNumberOfOutputs { num_outputs } => return _outputs.len() == num_outputs as usize,
```
mattdean-digicatapult, line 39: `_outputs` is now used so should be renamed to `outputs`. Just a rust lang convention

pallets/process-validation/src/restrictions.rs
```diff
@@ -219,4 +221,54 @@ mod tests {
...
+            &_outputs,
+        );
+        assert!(result);
+    }
+
+    #[test]
+    fn fixed_number_of_output_restriction_matches_fixed_output_total_fail() {
+        let mut is_owner: BTreeMap<u32, u64> = BTreeMap::new();
+        is_owner.insert(Default::default(), 1u64);
+        let _outputs = vec![
```
mattdean-digicatapult, line 254: rename

mattdean-digicatapult (changes requested):
Very close, just a few renames

mattdean-digicatapult (approved):

PR #54, opened 2022-03-02
Restrict a metadata value for an output to a fixed value

jonmattgray:
Allow restriction on the value of a metadata item for an **output** at a specific index in the output `Vec`. This restriction is limited to a pre-specified value and cannot depend on other inputs or outputs.

In polkadot apps:
image


README.md
```diff
@@ -172,7 +173,7 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
     "num_outputs": "u32"
   },
   "FixedMetadataValueRestriction": {
```
jonmattgray, line 175: ok to reuse this for two different restrictions?

mattdean-digicatapult (approved):

README.md
mattdean-digicatapult: yep that's fine

PR #56, opened 2022-03-30
Restrict role a transaction executor must have on an input

jonmattgray:
Allow restriction on the identity of a named role on an input to the sender of a run_process transaction

In polkadot apps:
image

### DIFF
--- a/pallets/process-validation/src/restrictions.rs
+++ b/pallets/process-validation/src/restrictions.rs
@@ -86,14 +86,10 @@ where
         }
         Restriction::SenderHasInputRole { index, role_key } => {
             let selected_input = &inputs[index as usize];
-            let sender_has_role = match selected_input.roles.get(&role_key) {
+            match selected_input.roles.get(&role_key) {
                 Some(account) => sender == account,
                 None => false,
-            };
-            if !sender_has_role {
-                return false;
             }
-            true
         }
         Restriction::SenderOwnsAllInputs => {
             for input in inputs {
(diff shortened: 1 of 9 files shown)


pallets/process-validation/src/restrictions.rs
```diff
@@ -77,6 +84,17 @@ where
...
+        Restriction::SenderHasInputRole { index, role_key } => {
+            let selected_input = &inputs[index as usize];
+            let sender_has_role = match selected_input.roles.get(&role_key) {
+                Some(account) => sender == account,
+                None => false,
+            };
+            if !sender_has_role {
+                return false;
+            }
+            true
```
mattdean-digicatapult, line 96: this can be simplified I think to:

```rust
match selected_input.roles.get(&role_key) {
   Some(account) => sender == account,
   None => false,
 }
```

runtime/src/lib.rs
```diff
@@ -93,7 +93,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("vitalam-node"),
     impl_name: create_runtime_str!("vitalam-node"),
     authoring_version: 1,
-    spec_version: 270,
```
mattdean-digicatapult, line 96: we haven't deployed this yet so I think on this occasion we can stick to `270`. Need to work out a runtime versioning system

pallets/process-validation/src/restrictions.rs
```diff
@@ -608,4 +626,72 @@ mod tests {
...
+        let roles = BTreeMap::from_iter(vec![(Default::default(), 1)]);
+        let inputs = vec![ProcessIO {
+            roles: roles.clone(),
+            metadata: BTreeMap::new(),
+            parent_index: None,
+        }];
+        let result = validate_restriction::<u64, u32, u32, u64>(
+            Restriction::SenderHasInputRole {
+                index: 0,
+                role_key: Default::default(),
```
mattdean-digicatapult, line 641: you asked before about other role_keys. Note that the `R` in `validate_restriction` is a `u32` so that's your `role_key` type

mattdean-digicatapult (changes requested):

runtime/src/lib.rs
jonmattgray: rolled back to `270`

pallets/process-validation/src/restrictions.rs
jonmattgray: done

pallets/process-validation/src/restrictions.rs
jonmattgray: thanks, I've added a test to check failing on an incorrect role (that isn't the default role)

mattdean-digicatapult (approved):

PR #57, opened 2022-04-04
Rename to DSCP

jonmattgray:
PR designed to merge shortly before/after renaming the repo to `dscp-node`

Bump to `v3.0.0`
`vitalam-pallet-traits` -> `dscp-pallet-traits` + all related pallets bumped a major version.


runtime/Cargo.toml
```diff
@@ -2,8 +2,8 @@
 authors = ['Digital Catapult <https://www.digicatapult.org.uk>']
 edition = '2018'
 license = 'Apache-2.0'
-repository = 'https://github.com/digicatapult/vitalam-node/'
-name = 'vitalam-node-runtime'
+repository = 'https://github.com/digicatapult/dscp-node/'
+name = 'dscp-node-runtime'
 version = '2.8.0'
```
mattdean-digicatapult, line 7: runtime version should change to 300 in `lib.rs` and match both here and in the node which references as a dependency

mattdean-digicatapult (commented):
Need to update spec version and runtime package version everywhere

mattdean-digicatapult (approved):

PR #59, opened 2022-04-11
Publish API package

jonmattgray:
Add `npm` package for a thin polkadot.js API that includes the additional types used in `dscp-node`. To be imported by other repos that interact with the node, such as `dscp-api`, so types don't have to be updated in multiple places for every change to the substrate node.

- [x] new package with boilerplate
- [x] workflows for linting + testing (against the latest node) + releasing the package 
- [x] script to check the package version matches the substrate runtime version


dscp-node/src/index.js
```diff
@@ -0,0 +1,82 @@
...
+  FixedMetadataValueRestriction: {
+    index: 'u32',
+    metadata_key: 'TokenMetadataKey',
+    metadata_value: 'TokenMetadataValue',
+  },
+  IsNew: 'bool',
+  Restrictions: 'Vec<Restriction>',
+}
+
+module.exports = types
```
mattdean-digicatapult, line 82: I do think the types should be exported bt we should also export the built `@polkadot-js/api` as well. This will like need to be wrapped in function to build the api so appropriate parameters (node location etc) can be passed in.

dscp-node/test/helper/logger.js
```diff
@@ -0,0 +1,21 @@
...
+        } else {
+          return '[REDACTED]'
+        }
+      },
+    },
+  },
+  process.stdout
+)
+
+module.exports = logger
```
mattdean-digicatapult, line 21: why do we need a mock logger?

mattdean-digicatapult (changes requested):
We're getting there. This first set of comments is mostly focussed on the shape of the package. We should also talk about test methodology because I'm not totally convinced there (but can't necessarily think of a better approach)

dscp-node/test/helper/logger.js
jonmattgray: to log during tests. If we're moving to exporting the whole api though I'll add the logger to `src/` and have its log level configurable as well

dscp-node/src/env.js
```diff
@@ -0,0 +1,17 @@
...
+  dotenv.config({ path: path.resolve(__dirname, '../.env') })
+}
+
+const vars = envalid.cleanEnv(process.env, {
+  METADATA_KEY_LENGTH: envalid.num({ default: 32 }),
+  METADATA_VALUE_LITERAL_LENGTH: envalid.num({ default: 32 }),
+  PROCESS_IDENTIFIER_LENGTH: envalid.num({ default: 32 }),
+})
+
+module.exports = vars
```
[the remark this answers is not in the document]
jonmattgray: changed so the package exports `buildApi` which takes an `options` object for these arguments

dscp-node/src/index.js
jonmattgray: now exports `api, types, keyring`

api/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+    IsNew: 'bool',
+    Restrictions: 'Vec<Restriction>',
+  }
+
+  const apiOptions = {
+    provider,
+    types,
+  }
+
+  const api = new ApiPromise(apiOptions)
```
n3op2, line 99: like that we include a connection (api object), gives us a chance to alter polkadot api object

api/lib/api.js
```diff
@@ -0,0 +1,122 @@
...
+  api.on('disconnected', () => {
+    logger.warn(`Disconnected from substrate node at ${options.apiHost}:${options.apiPort}`)
+  })
+
+  api.on('connected', () => {
+    logger.info(`Connected to substrate node at ${options.apiHost}:${options.apiPort}`)
+  })
+
+  api.on('error', (err) => {
+    logger.error(`Error from substrate node connection. Error was ${err.message || JSON.stringify(err)}`)
```
n3op2, line 110: @mattdean-digicatapult how about if this could take a function? or some args to forward to the server or anything else really

api/test/helper/routes.js
```diff
@@ -0,0 +1,71 @@
+const { buildApi } = require('../../lib')
```
n3op2, line 1: So we are going beyond just types publishing the entire package? There will be a few changes required across other repos, but like the idea of having to maintain this, but then might make sense to create it as separate repo and import from repo?

api/test/integration/types.test.js
```diff
@@ -0,0 +1,34 @@
+const { describe, test } = require('mocha')
+const { expect } = require('chai')
+
+const { setupApi, runProcess, utf8ToUint8Array, getLastTokenId, getToken } = require('../helper/routes')
+
+const USER_ALICE_TOKEN = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'
+const METADATA_KEY_LENGTH = 32
+const METADATA_VALUE_LITERAL_LENGTH = 32
+
+describe('types', function () {
```
n3op2, line 10: knowing that we are using all method e.g. getToken, runProcess and so on it would be nice to get some coverage on that

api/test/integration/types.test.js
```diff
@@ -0,0 +1,34 @@
...
+  const context = {}
+
+  before(async function () {
+    await setupApi(context)
+  })
+
+  describe('health check', function () {
+    test('mint token', async function () {
+      const lastTokenId = await getLastTokenId(context)
```
n3op2, line 19: you can also wrap 

```js
const lastTokenId = await getLastTokenId(context)
```

in the below. It also might be a good idea to assert for more props as at the moment we just check for id

```js
befooreEach(async () => {
  await getLastTokenId(context)
})
```

api/test/integration/types.test.js
```diff
@@ -0,0 +1,34 @@
...
+      const roles = new Map(Object.entries({ 0: USER_ALICE_TOKEN }))
```
n3op2, line 20: This could be exported as fixture and used across all tests or some helper method for different scenarios e.g. errors

api/test/helper/routes.js
mattdean-digicatapult: The huge benefit of doing it here is that we can tie changes to api library to the runtime changes. That said this should be a pretty thin cover

api/test/integration/types.test.js
jonmattgray: I see the tests as more asserting that we get 'something' rather than the props of that thing. If the types are wrong the whole thing blows up with a decoding error

.github/workflows/test-types.yml
```diff
@@ -0,0 +1,87 @@
+name: Test types
+
+on:
+  push:
+    paths: 
+      - 'runtime/**'
+      - 'dscp-node/test/**'
```
mattdean-digicatapult, line 7: update path (should be `api`)

api/lib/api.js
mattdean-digicatapult, line 110: I htink these behaviours are perhaps a touch over-opinionated. aybe remove theese three listenners which will also remove the need to have a `logger` instance

mattdean-digicatapult (changes requested):
Couple of small changes

mattdean-digicatapult:
@jonmattgray can we also add

```
api.isReadyOrError.catch(() => {})
```

to the API after `const api = new ApiPromise(apiOptions)` so that the undandled promise rejection doesn't happen?

.github/workflows/test-types.yml
jonmattgray: thanks

api/lib/api.js
jonmattgray: removed `logger`

mattdean-digicatapult (approved):

PR #60, opened 2022-04-14
Restrict role a transaction executor must have on an output

jonmattgray:
Allow restriction on the identity of a named role on an output to the sender of a run_process transaction

image

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -4,7 +4,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/dscp-node/'
 name = 'dscp-node-runtime'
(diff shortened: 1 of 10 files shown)


runtime/Cargo.toml
```diff
@@ -4,7 +4,7 @@ edition = '2018'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/dscp-node/'
 name = 'dscp-node-runtime'
-version = '3.0.1'
+version = '3.1.0'
```
mattdean-digicatapult, line 7: we should probably bump the version in lib.rs as well. Just because we have now redoployed the cluster so the current version is that in staging

mattdean-digicatapult (approved):

PR #63, opened 2022-04-19
Restrict output metadata value type

mattdean-digicatapult:
Allows restricting the type of a metadata value for a specific output key on a specific output index

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,9 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
       "None": null
     }
   },
+  "MetadataValueType": {
+      "_enum": ["File", "Literal", "TokenId", "None"],
+  },
   "Role": {
     "_enum": [
       "Owner",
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -192,6 +196,11 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
     "metadata_key": "TokenMetadataKey",
     "metadata_value": "TokenMetadataValue"
   },
+  "FixedOutputMetadataValueType": {
```
jonmattgray, line 199: should have restriction on the end? `FixedOutputMetadataValueTypeRestriction`

mattdean-digicatapult: renamed

pallets/process-validation/src/restrictions.rs
```diff
@@ -44,22 +45,28 @@ where
         metadata_key: TokenMetadataKey,
         metadata_value: TokenMetadataValue,
     },
+    FixedOutputMetadataValueType {
+        index: u32,
+        metadata_key: TokenMetadataKey,
+        metadata_value_type: TokenMetadataValueDiscriminator,
```
jonmattgray, line 51: does `TokenMetadataValueDiscriminator` end up as dropdown of types in the polkadot explorer?

mattdean-digicatapult: I don't believe so. I believe the frontend uses the supplied types to generate names. These names are all internal to the build

jonmattgray (approved):

PR #66, opened 2022-04-25
Add boolean binary restriction

jonmattgray:
Boolean logic (AND/OR/XOR/NOR/NAND) for restrictions so you can logically combine different restrictions together.

- [x] new `Combined` restriction
- [x] `create_process` checks depth of restriction. Maximum set to 3 for runtime.
- [x] tests for restriction and depth check  

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
   "Restriction": {
     "_enum": {
       "None": "()",
-      "BooleanBinary": "BooleanBinaryRestriction",
+      "Combined": "CombinedRestriction",
       "SenderOwnsAllInputs": "()",
       "SenderHasInputRole": "SenderHasInputRoleRestriction",
       "SenderHasOutputRole": "SenderHasOutputRoleRestriction",
@@ -179,7 +179,7 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
       "FixedOutputMetadataValueType": "FixedMetadataTypeRestriction"
     }
   },
-  "BooleanBinaryRestriction": {
+  "CombinedRestriction": {
     "operator": "BinaryOperator",
     "restriction_a": "Box<Restriction>",
     "restriction_b": "Box<Restriction>"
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -162,6 +165,7 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
   "Restriction": {
     "_enum": {
       "None": "()",
+      "BooleanBinary": "BooleanBinaryRestriction",
```
jonmattgray, line 168: not sure about `BooleanBinary` as a name?

mattdean-digicatapult: `CombinedRestriction`?

README.md
```diff
@@ -175,6 +179,11 @@ In order to use the API within `polkadot.js` you'll need to configure the follow
       "FixedOutputMetadataValueType": "FixedMetadataTypeRestriction"
     }
   },
+  "BooleanBinaryRestriction": {
+    "operator": "BinaryOperator",
+    "restriction_a": "Box<Restriction>",
+    "restriction_b": "Box<Restriction>"
```
mattdean-digicatapult, line 185: They need to be declared as boxes in the types?

pallets/process-validation/src/lib.rs
```diff
@@ -193,6 +206,34 @@ pub mod pallet {
...
+                return true;
+            }
+
+            match restriction {
+                Restriction::BooleanBinary {
+                    operator: _,
+                    restriction_a,
+                    restriction_b
+                } => {
+                    count += 1;
```
mattdean-digicatapult, line 229: instead of having a mutible `count`it might be better to assign a new variable here to `count + 1`.

pallets/process-validation/src/restrictions.rs
```diff
@@ -1393,4 +1439,292 @@ mod tests {
...
+            },
+            &1,
+            &Vec::new(),
+            &outputs
+        );
+        assert!(!result);
+    }
+
+    #[test]
+    fn boolean_binary_or_succeeds() {
```
mattdean-digicatapult, line 1508: I think we're missing an `or` failure case

README.md
mattdean-digicatapult: Actually yes I understand this needs to be boxed here

pallets/process-validation/src/lib.rs
jonmattgray: thanks, removed the `clone()`s too

README.md
jonmattgray: changed to `Combined` and `CombinedRestriction`

mattdean-digicatapult (approved):

PR #67, opened 2022-06-13
Feature/voting

mattdean-digicatapult:
Adds voting support to the runtime. The thresholds for voting are somewhat arbtrary at the moment by highlight the art of the possible. Notably:

1. Root like operations require a majority of the `TechnicalCommittee`
2. Rotating the IPFS key should always be safe so only requires 2 members of the `TechnicalCommittee`
3. Process flow changes are safe so only require 2 members of the `TechnicalCommittee`

`Sudo` is still present and can do everything as before

### Implementation notes

The `pallet-collective` only allows us to generate new `Origin`s that can be used when restricting call permissions. In order to perform opperations that would usually require `Root` permissions (such as `system::set_code` we need a way of enabling the collective to perform `Root` operations. The way I've done this is by implementing a new pallet `pallet-doas` (named for the sudo alternative [doas](https://man.openbsd.org/doas)). This allows us to configure an Origin in the runtime that will be able to act like `Root`. We then configure that Origin to be sourced from the `TechnicalCommittee` implemented using `pallet-collective`

### DIFF
--- a/pallets/doas/src/lib.rs
+++ b/pallets/doas/src/lib.rs
@@ -44,9 +44,9 @@ pub mod pallet {
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
         /// A doas_root just took place. \[result\]
-        DoasRootDone(DispatchResult),
+        DidAsRoot(DispatchResult),
         /// A doas just took place. \[result\]
-        DoasDone(DispatchResult),
+        DidAs(DispatchResult),
     }
 
     // The pallet's dispatchable functions.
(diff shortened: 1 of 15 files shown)


pallets/doas/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+          let dispatch_info = call.get_dispatch_info();
+          (
+            dispatch_info.weight
+              .saturating_add(10_000)
+              // AccountData for inner call origin accountdata.
+              .saturating_add(T::DbWeight::get().reads_writes(1, 1)),
+            dispatch_info.class,
+          )
+        })]
+        pub(super) fn doas(
```
n3op2, line 128: looks straight forward

n3op2 (approved):

jonmattgray (approved):

PR #80, opened 2022-07-08
Remove types from API and upgrade polkadot

mattdean-digicatapult:
Upgrades the API and removes the unnecessary type annotations.


jonmattgray:
Could remove types from the README?

jonmattgray (approved):

PR #81, opened 2022-07-14
Added ability to publish to npmjs

dblane-digicatapult:
Added a secondary publish job which publishes to npm registry.

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -9,5 +9,5 @@ if [ -z $CI_PROJECT_NAME ] ; then
    rustup update stable
 fi
 
-rustup target add wasm32-unknown-unknown --toolchain nightly-2021-11-09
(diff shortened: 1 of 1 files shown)


node/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = '2021'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/dscp-node/'
 name = 'dscp-node'
-version = '4.0.0'
```
mattdean-digicatapult, line 9: need to also update the version of the dependency `dscp-node-runtime` in this file and then regenerate the cargo lock

mattdean-digicatapult (changes requested):

mattdean-digicatapult (approved):

PR #85, opened 2022-09-26
Increase max metadata count to 64

jonmattgray:
Increases the maximum allowed metadata count on a token from `16` to `64`


dblane-digicatapult (approved):
LGTM.

PR #86, opened 2022-10-14
Re-enable combined restriction functionality

mattdean-digicatapult:
Re-enables the functionality that used to be encoded in the Combined restriction type. To avoid circular type references and to simplify the API this has instead been achieved by abstracting out the way restricitons are combined. Previously top-level we assumed that restrictions should be 'AND'ed together. Now an arbitrary binary expression tree is used  written in postfix notation.

Also added are a Fail restriction type which always fails to pass. This is useful both for testing and for doing certain unary operations as the tree only excepts binary operations.

There is some opinionation as to what the defaul should be for the Process program. This change maintains current behaviour that the program will always succeed.

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@digicatapult/dscp-node",
-  "version": "4.2.2",
+  "version": "4.3.2",
   "description": "DSCP node types",
   "repository": {
     "type": "git",
--- a/api/package.json
+++ b/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@digicatapult/dscp-node",
-  "version": "4.3.2",
+  "version": "4.3.0",
   "description": "DSCP node types",
   "repository": {
     "type": "git",
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -6,7 +6,7 @@ edition = '2021'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/dscp-node/'
 name = 'dscp-node'
-version = '4.3.2'
+version = '4.3.0'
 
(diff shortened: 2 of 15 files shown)


api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@digicatapult/dscp-node",
-  "version": "4.2.2",
+  "version": "4.3.2",
```
dblane-digicatapult, line 3: 4.3.0

node/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = '2021'
 license = 'Apache-2.0'
 repository = 'https://github.com/digicatapult/dscp-node/'
 name = 'dscp-node'
-version = '4.2.2'
+version = '4.3.2'
```
dblane-digicatapult, line 9: 4.3.0

node/Cargo.toml
```diff
@@ -55,7 +55,7 @@ frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/parityte
 frame-benchmarking-cli = { version = "4.0.0-dev", git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.25" }
 
 # Local Dependencies
-dscp-node-runtime = { path = '../runtime', version = '4.2.2' }
+dscp-node-runtime = { path = '../runtime', version = '4.3.2' }
```
dblane-digicatapult, line 58: again.

dblane-digicatapult (changes requested):
bump the runtime spec version

dblane-digicatapult (approved):

jonmattgray (approved):

PR #93, opened 2022-10-26
Add benchmarks and weights for pallet-process-validation

jonmattgray:
Benchmarks for:

1. `create_process`
    - iterate over number of program input
    - maximum length is ~~200~~ 201 of restrictions + boolean ops (length is always odd)

2. `disable_process`

Weights generated and attatched to extrinsics

### DIFF
--- a/pallets/process-validation/src/benchmarking.rs
+++ b/pallets/process-validation/src/benchmarking.rs
@@ -9,6 +9,7 @@ use crate::Pallet as ProcessValidation;
 
 benchmarks! {
   create_process {
+    // valid programs have x Restrictions and (x-1) Ops, therefore number of BooleanExpressionSymbol to add is always odd
     let r in 1 .. (1 + T::MaxProcessProgramLength::get() / 2);
 
     let mut program = BoundedVec::<_, _>::with_bounded_capacity(T::MaxProcessProgramLength::get() as usize);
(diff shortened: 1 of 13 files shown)


runtime/Cargo.toml
```diff
@@ -118,27 +122,28 @@ runtime-benchmarks = [
     "pallet-scheduler/runtime-benchmarks",
     "pallet-simple-nft/runtime-benchmarks",
     "pallet-symmetric-key/runtime-benchmarks",
+    "pallet-process-validation/runtime-benchmarks",
     "sp-runtime/runtime-benchmarks",
 ]
 try-runtime = [
-	"frame-try-runtime",
```
jonmattgray, line 124: this churn is tabs -> spaces

pallets/process-validation/src/benchmarking.rs
```diff
@@ -1,25 +1,51 @@
...
+      }
+    }
+
+  }: _(RawOrigin::Root,
+  T::ProcessIdentifier::default(),
+  program.clone())
   verify {
+    // let version = ProcessValidation::<T>::get_version(&T::ProcessIdentifier::default());
+    // let process = ProcessModel::<T>::get(T::ProcessIdentifier::default(), version);
+    // assert_eq!(process.status, ProcessStatus::Enabled)
```
jonmattgray, line 32: not found a way to get this passing correctly. Also tried checking the `i`th version is enabled but the process isn't incrementing its version as expected.

jonmattgray: any suggestions welcome 🙏

mattdean-digicatapult: hmmm, I'd have to look into it more carefully. We should get this working though before merging

mattdean-digicatapult (approved):
lgtm

PR #97, opened 2022-10-28
Automatically generating weights

dblane-digicatapult:


README.md
```diff
@@ -74,20 +74,17 @@ For `dev` chain, the network only contains a node for `Alice` so other nodes wil
...
 
 ```bash
-cargo build --release --features runtime-benchmarks
+cargo run --release --features=runtime-benchmarks \
+--manifest-path=Cargo.toml \
+-- benchmark pallet --chain=dev \
+--steps=50 --repeat=20 \
+--pallet=pallet-symmetric-key \
+--extrinsic=* --execution=wasm --wasm-execution=compiled \
+--heap-pages=4096 --output=./pallets/symmetric-key/src/weights.rs \
```
jonmattgray, line 83: how do we run it for all pallets?

mattdean-digicatapult: it wouldn't be hard to script. The bigger issue is each pallet will need a different value for `steps` and `repeat`

dblane-digicatapult: The biggest issue in my opinion is that we don't necessarily want to be doing this in CI as there are no performance guarantees between nodes.

dblane-digicatapult: If we were to do it, we could simply create a matrix job to perform it.

mattdean-digicatapult: maybe the risk being forgetting to add something to the matrix

mattdean-digicatapult: do we know what best practices are for doing benchmarking in terms of hardware?

dblane-digicatapult: No, but...
https://github.com/paritytech/substrate/blob/master/utils/frame/benchmarking-cli/src/machine/README.md

dblane-digicatapult: We should also probably use `--production` instead of `--release` for benchmarks (and for artifacts)

PR #101, opened 2022-11-08
Add Error To PR when Pallets with weights are changed

nitro-marky:
If a pallet containing weight is edited a comment is added to the PR to check that the weights haven't been affected.

### DIFF
--- a/node_modules/.package-lock.json
+++ b/node_modules/.package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "dscp-node",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "dscp-node",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dscp-node-runtime"
-version = "4.4.3"
+version = "4.4.4"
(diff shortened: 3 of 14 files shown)


runtime/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "dscp-node-runtime"
-version = "4.4.3"
+version = "4.4.4"
```
mattdean-digicatapult, line 3: @dblane-digicatapult are we keeping the `dscp-node` version is lockstep with the `dscp-runtime` these days? I ask because if it is we need to also bump the `spec_version` and bump the referenced version in the `dscp-node` `Cargo.toml`

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
jonmattgray, line 1: this one too, can be deleted?

mattdean-digicatapult: yes delete

node_modules/.package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
mattdean-digicatapult: and yes delete

runtime/Cargo.toml
mattdean-digicatapult: ok so @nitro-marky we need to update the `spec_version` in the runtime `lib.rs` to be `444` and we need to update the version of the `dscp-runtime` referenced in the `node/Cargo.toml` to be `4.4.4`

dblane-digicatapult (approved):
LGTM.

mattdean-digicatapult (approved):

PR #109, opened 2022-11-24
Improve schedule tests on symmetric key

mattdean-digicatapult:
The current test logic on `pallet-symmetric-key`'s `schedule` tests is a bit flawed. The intension is to assert that the schedule indeed runs automatically on an interval but the tests don’t actually do this. Instead what’s happening is the schedule is initialised once (and this is tested), but the following interval is not. This ticket will look to improve the tests by

- making explicit that the existing test is only testing the schedule initialisation
- adding another test to assert that the schedule indeed fires on an interval

To achieve the latter case I had to implement our own `TestRandomness` as the one in substrate doesn’t actually change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1347,7 +1347,7 @@ checksum = "9ea835d29036a4087793836fa931b08837ad5e957da9e23886b29586fb9b6650"
 
 [[package]]
 name = "dscp-node"
-version = "4.4.7"
+version = "4.4.8"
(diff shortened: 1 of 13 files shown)


mattdean-digicatapult:
This change doesn't affect any of the pallet extrinsics in a meaningful way so shouldn't affect wieghts

jonmattgray (approved):

PR #117, opened 2023-03-15
Adds MatchInputIdOutputMetadataValue restriction

mattdean-digicatapult:
Enforce that a metadata value on a specified metadata key on an output at a specified index matches the id of an input at a specified index.

For L3 this will allow us to enforce that on final acceptance a match can only finalise for the correct demands.

Note this also deals with the index out-of-bounds bug but only for this new Restriction


dblane-digicatapult (approved):

jonmattgray (approved):

n3op2 (approved):
might not be the person to review (need a refreshment)

PR #118, opened 2023-03-15
Chore/simplify simple nft

mattdean-digicatapult:
Simplify the simple nft pallet to remove the possibility of not using a process flow. This is now mandatory. Also removes `SenderOwnsAllInputs` restriction. This will be a major version bump.

Checklist:

- [x] remove `SenderOwnsAllInputs`
- [x] remove possiblity of not passing a process flow from runProcess
- [x] benchmarks for simple nft pallet
- [x] benchmarks for simple nft pallet with correct restrictions


jonmattgray (approved):

rust-toolchain
```diff
@@ -1 +1 @@
-nightly
+nightly-2023-03-06
```
dblane-digicatapult, line 1: Nightly broke?

mattdean-digicatapult: yep

PR #119, opened 2023-03-17
Restriction out-of-bounds checks

mattdean-digicatapult:
Adds checks for restrictions accessing inputs/outputs that are out of bounds. Prevents panics at runtime.

Also adds 3 new restrictions so we can cover original-token-id behaviour in full

Also adds renovate config at @dblane-digicatapult's request


pallets/process-validation/src/restrictions.rs
```diff
@@ -26,6 +26,18 @@ pub enum Restriction<RoleKey, TokenMetadataKey, TokenMetadataValue, TokenMetadat
...
     },
+    OutputHasMetadataValue {
+        index: u32,
+        metadata_key: TokenMetadataKey
+    },
+    InputHasRole {
+        index: u32,
+        role_key: RoleKey
+    },
+    InputHasMetadataValue {
```
jonmattgray, line 37: should `HasMetadataValue` be `HasMetadataKey`?

mattdean-digicatapult: 🤔 , maybe `HasMetadata`? Because it's whether it has that metadata entry

jonmattgray: yeah `HasMetadata` would match `HasRole`

jonmattgray: in that they have a similar name and are both checking there is a role/metadata entry with any value

pallets/process-validation/src/restrictions.rs
```diff
@@ -26,6 +26,18 @@ pub enum Restriction<RoleKey, TokenMetadataKey, TokenMetadataValue, TokenMetadat
         index: u32,
         role_key: RoleKey
     },
+    OutputHasMetadataValue {
```
jonmattgray, line 29: new restrictions can be added to the README

jonmattgray (approved):

PR #130, opened 2023-04-18
Switch to babe consensus

jonmattgray:
Switch from `aura` -> `babe` consensus
Switch from `type Randomness = RandomnessCollectiveFlip` -> `type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>`

Example contents from the `local_testnet` spec:
```
      "babe": {
        "authorities": [
          [
            "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
            1
          ],
          [
            "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
            1
          ],
          [
            "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
            1
          ]
        ],
        "epochConfig": {
          "c": [
            1,
            4
          ],
          "allowed_slots": "PrimaryAndSecondaryPlainSlots"
        }
```
What is `c`?
A constant value that is used in the threshold calculation formula. Expressed as a rational where the first member of the tuple is the numerator and the second is the denominator. The rational should represent a value between 0 and 1. In the threshold formula calculation, 1 - c represents the probability of a slot being empty.
https://paritytech.github.io/substrate/master/sp_consensus_babe/struct.BabeEpochConfiguration.html


dblane-digicatapult (approved):
Node successfully starts with a new genesis generated by our `make-new-cluster-genesis` script.

runtime/Cargo.toml
mattdean-digicatapult, line 1: we should be able to remove `pallet-insecure-randomness-collective-flip` now right?

jonmattgray: yes thanks, I missed that. Surprised cargo didn't complain, need a `cargo depcheck`...

mattdean-digicatapult (approved):

PR #137, opened 2023-07-17
switch to BABE consensus

kalaninja:
- switch from AURA to BABE consensus
- increment major version
- use BABE rpc extensions, mandatory for 0.9.42 (https://github.com/paritytech/substrate/pull/13883)


mattdean-digicatapult:
Thanks so much for your help with this @kalaninja! Because this is quite a major change to the runtime/node we will be doing a bit of manual testing (@jonmattgray I beleive you'll be looking at this) just to make sure it works as we expect. Hopefully shouldn't take too long, but it might take a bit longer to merge than usual.

Once again thank you for the contribution!

jonmattgray (approved):
LGTM. Tested with 3 nodes, which successfully peered and finalised blocks with no errors. We can replace use of `pallet_insecure_randomness_collective_flip` with babe's randomness for `pallet_symmetric_key` in a future PR

mattdean-digicatapult (commented):
I think the main remaining question I have left is is around upgrade path. Do we know how/if we can upgrade the deployed l3 network with this change? 

We do have the option in the worst case of just restarting the chain with a fresh genesis but that's clearly not ideal.

mattdean-digicatapult:
Closed in favour of #140 which includes this work. Thanks @kalaninja !

PR #142, opened 2023-09-08
Feature/l3 168

Ellenn-A:
Hello guys, 

Please have a look and let me know if you find any issues. I have added some questions as there are some additional actions I think I should do, but I want to make sure I have not misunderstood.

in PR: 
- added profile.production
- regenerated pallet & storage benchmarks and integrated them with runtime (benchmarks ran on AWS node)

Questions: 
- can I update the readme so instead of using ```cargo build --release``` we will use ```cargo build --profile production```? 
(can I update that throughout the read me?)
- because ```cargo build --profile production``` produces ./target/production/dscp-node  can I change references in the README from e.g. ```./target/release/dscp-node --dev``` to ```./target/production/dscp-node --dev``` (this is an example of running dev-chain node)
- the last point of my ticket: ```update workflows to reflect new build arguments``` -> can I change all references in release.yml  from ```./target/*/release``` to ```./target/*/production```? (I don't think it's used in the check-and-test.yml?)


mattdean-digicatapult (changes requested):
This looks really close. Couple of things to do I think:

* indentation seems to have gone wonky. Your changes all indent with tabs when the rest of the repo is spaces. We should make your changes consistne twith the rest of the repo.
* Need to update the build of the release artifact in https://github.com/digicatapult/dscp-node/blob/a7eabdb5c267bf598b0a1d9c9805fd0a96a07ab5/.github/workflows/release.yml#L119

mattdean-digicatapult (approved):

mattdean-digicatapult (approved):

PR #143, opened 2023-09-12
delete_token extrinsic and token cleanup

mattdean-digicatapult:
Implements `delete_token` extrinsic and the cleanup logic that should run `on_idle`. Runtime sets the tombstoning period as 7 days. 

This PR includes additional tests for the new extrinsic and `on_idle` behaviour and weights have been updated on a `t3a.2xlarge` instance in AWS


n3op2 (approved):

n3op2:
might be a good idea to get a second pair of eyes. Did not spot anything obvious, like the idea of not having to hold that many active tokens knowing that we do produce a fair number. Approved

pallets/utxo-nft/src/lib.rs
```diff
@@ -168,27 +181,84 @@ pub mod pallet {
...
+                    .checked_div_per_component(
+                        // for each delete operation we fetch the graveyard entry, delete the token, then delete the graveyard entry
+                        &T::WeightInfo::delete_token()
+                            .saturating_add(T::DbWeight::get().reads(1))
+                            .saturating_add(T::DbWeight::get().writes(1))
+                    )
+                    .unwrap_or(0),
+                None => 0
+            };
+
```
mattdean-digicatapult, line 226: return if 0

pallets/utxo-nft/src/lib.rs
```diff
@@ -168,27 +181,84 @@ pub mod pallet {
...
+                })
+                .unwrap_or((iter_count, iter_count));
+
+            // write graveyard state (base_weight)
+            <CurrentGraveyardState<T>>::put(GraveyardState {
+                start_index: start_index + delete_count,
+                end_index
+            });
+
+            let spent_weight = base_weight.saturating_add(T::WeightInfo::delete_token().mul(delete_op_count));
```
mattdean-digicatapult, line 259: doesn't take into account extra db read and write per loop

pallets/utxo-nft/src/weights.rs
mattdean-digicatapult, line 1: need to update benchmarks for run_process

n3op2 (approved):

PR #145, opened 2023-10-02
DSCP token lang MVP

mattdean-digicatapult:
MVP of a language definition and parser for building DSCP token models.

Todo:

- [x] language syntax rules
- [x] Translate l3 token model
- [x] AST building
- [x] Integration tests for parsing string to ast
- [x] unit testing for some parse functions
- [x] cli interface for parsing an input file
- [x] Documentation


mattdean-digicatapult:
I'm marking this as ready for review though more is needed to get this complete. Mostly the need is for documentation which we need to agree how to split between here and the `dscp-documentation` repo. I'd like this to be reviewd in the context of the ticket `l3-213 ` for now which is to just produce a grammar and parser that passes and fails based on validity of the input to the grammar. I think the tests I now have are sufficient for that.

As part of doing this I had to make some unrelated changes to the repo. Notably:

- switching toolchain to `stable`. To use the latest version of `clap` to pass cli args I had to upgrade the toolchain and stable works now 🎉 
- removing `try-runtime` as a feature of the node. Looking at `substrate` they are moving towards try-runtime being done by a separate executable to the node. As `try-runtime` never worked for us this seems a small loss and it was now causing compile errors
-  formatting changes are present due to the switch to `stable`. Sorry

The easiest way to test this is to use the example token model which is based off of l3. You can check this with 

```
cargo run -p dscp-lang -- parse -v ./tools/lang/examples/l3.dscp
```

Where pasing `-v` leads to more verbose output.

Feel free to edit the model and play, this generates quite nice errors like: 

```
Error occurred when parsing grammar:   --> 39:15
   |
39 |   a.optimiser = optimiser,
   |               ^---
   |
   = expected eq, neq, is, or isnt
```

tools/lang/examples/l3.dscp
```diff
@@ -0,0 +1,192 @@
...
+  a == b,
+  a.optimiser == b.optimiser,
+  a.member_a == b.member_a,
+  a.demand_a == b.demand_a,
+  a.member_b == b.member_b,
+  a.demand_b == b.demand_b,
+  a.replaces == b.replaces,
+}
+
+fn clone_match | foo: Match2 | => | bar: Match2 | where {
```
Ellenn-A, line 47: did you want to switch this back to 'a' and 'b' instead of 'foo' and 'bar'? Is just a cosmetic issue

mattdean-digicatapult: yep good spot

Ellenn-A (commented):
I would not rely on me when it comes to the types.rs and cli.rs, but I read through it

n3op2 (approved):

PR #151, opened 2023-12-18
Add option to enable manual sealing of blocks when in dev mode

mattdean-digicatapult:
When running integration tests on repositories that depend on `dscp-node` we have to limit tests that interact with the chain because of the 6s block generation time. This PR enables a `--manual_seal` option when running with `--dev` that will allow us to manually trigger block generation and finalisation usng an RPC command. This should make it much more practical for us to have an extensive suite of integration tests.


n3op2 (approved):

dblane-digicatapult (changes requested):
Should add some tests for this flag?

mattdean-digicatapult:
> Should add some tests for this flag?

I honestly don't see how @dblane-digicatapult without adding some substantial apparatus to the node. Given we don't test any of the rest of the CLI this doesn't seem like a regression.

dblane-digicatapult (approved):

PR #162, opened 2024-01-31
Add healthcheck to Dockerfile

mattdean-digicatapult:
Adds a healthcheck to the Dockerfile so that we can orchestrate the order in which services start in our docker compose files

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ FROM ubuntu:jammy AS runtime
 
 RUN <<EOF
 apt-get update
-apt-get install -y libgcc-11-dev
+apt-get install -y libgcc-11-dev curl
 rm -rf /var/lib/apt/lists/*
 EOF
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -53,4 +53,6 @@ WORKDIR /dscp-node
 
 ENTRYPOINT [ "/dscp-node/dscp-node" ]
 
-EXPOSE 30333 9933 9944
+HEALTHCHECK CMD curl -f http://localhost:9944/health || exit 1
+
+EXPOSE 30333 9615 9944
```
mattdean-digicatapult, line 58: Note this removes the old `ws` port that the node no longer listens on but documents the exposure of the prometheus telemetry port

n3op2: is anything else affected by this change or just prometheus?

n3op2 (approved):

PR #279, opened 2025-05-02
Add input references to runtime and run_prcess_as_root

mattdean-digicatapult:
# Pull Request

## Checklist
- [ ] Have you read Digital Catapult's [Code of Conduct](https://github.com/digicatapult/.github/blob/main/CODE_OF_CONDUCT.md)?
- [ ] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.

## PR Type

Please delete options that are irrelevant.

- [ ] Bug Fix
- [ ] Chore
- [ ] Feature
- [ ] Documentation Update
- [ ] Code style update (formatting, local variables)
- [ ] Breaking Change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/pallets/process-validation/src/migration.rs
+++ b/pallets/process-validation/src/migration.rs
@@ -1,0 +1,762 @@
+use super::*;
+use frame_support::traits::{OnRuntimeUpgrade, StorageVersion};
+use sp_runtime::Weight;
+
+/// The log target.
+const TARGET: &'static str = "runtime::symmetric-key::migration";
+
+pub mod v2 {
+    use super::*;
+
+    /// Migrate the symmetric-key pallet from V1 to V2.
+    pub struct MigrateToV2<T>(core::marker::PhantomData<T>);
+
+    impl<T: Config> OnRuntimeUpgrade for MigrateToV2<T> {
+        fn on_runtime_upgrade() -> Weight {
+            let version = StorageVersion::get::<Pallet<T>>();
+            if version != 1 {
+                log::warn!(
+                  target: TARGET,
+                  "skipping v1 to v2 migration: executed on wrong storage version. Expected version 1, found {:?}",
+                  version,
+                );
+                return T::DbWeight::get().reads(1);
+            }
+
+            Pallet::<T>::migrate_v1_to_v2() + T::DbWeight::get().reads(1)
+        }
+    }
+}
+
+impl<T: Config> Pallet<T> {
+    fn migrate_v1_to_v2() -> Weight {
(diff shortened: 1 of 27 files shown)


README.md
```diff
@@ -120,12 +120,25 @@ CurrentGraveyardState<T: Config> = StorageValue<_, GraveyardState, ValueQuery>;
 
 Tokens can be minted/burnt by calling the following extrinsic under `UtxoNFT`:
 
+TODO: fix docs
```
Ellenn-A, line 123: keeping this?

pallets/process-validation/src/migration.rs
```diff
@@ -0,0 +1,762 @@
...
+            } => Restriction::MatchArgsMetadataValue {
+                left_arg_type: ArgType::Input,
+                left_index: input_index,
+                left_metadata_key: input_metadata_key,
+                right_arg_type: ArgType::Output,
+                right_index: output_index,
+                right_metadata_key: output_metadata_key,
+            },
+            RestrictionOld::MatchInputIdOutputMetadataValue {
+                input_index,
```
Ellenn-A, line 195: this purposefully doesn't have  `input_metadata_key`, correct? because then it would have been the same restriction as `MatchInputOutputMetadataValue`

mattdean-digicatapult: So the id isn't stored in the input's metadata and never was. What this allows us to do is to set an output metadata value equal to an input token's id. We use this in the language to maintain the original id

tools/lang/src/compiler/mod.rs
```diff
@@ -165,6 +214,80 @@ fn make_process_restrictions(
...
+                .filter_map(|t| match &t.value {
+                    TokenFieldType::LiteralValue(v) => Some(v.value.to_owned()),
+                    _ => None,
+                })
+                .collect(),
+        })
+        .collect()
+}
+
+fn make_process_arguments(fn_decl: &FnDecl, token_decls: &HashMap<&str, TokenDecl>) -> Arguments {
```
mattdean-digicatapult, line 267: note to self this is broken. Inputs should be specified as token ids and for outputs we need to output the token type and version

Ellenn-A (commented):
read through, didn't notice anything obvious

jonmattgray (approved):

